PR #77, opened 2018-11-29
WIP: Connect the state machine with march_control

Ishadijcks:
A partial connection of the march_control package with the march_state_machine.

It is technically functional, but needs to be refactored before merging.

Features to add:
- [ ] Route gait requests through the march_api instead of directly to the march_control node
- [ ] The state machine should stay in idle and listen to (user) input on a specific topic


TimBuckers:
I think we need to redesign the state machine a bit, in response to the state machine meeting.

Ishadijcks:
Agreed

PR #177, opened 2019-03-06
Feature/update buildsh

TimBuckers:
Improve the build.sh script to check all packages. If a new march package is added this one is also automatically checked by the build.sh.

- pycodestyle . --max-line-length 120
- loop through all march directories on default
- Possible to make exceptions for a package for roslint

### DIFF
--- a/march_ws/build.sh
+++ b/march_ws/build.sh
@@ -63,7 +63,7 @@ do
     then
             package_path=$(dirname "${directory}")
             echo ${package_path}
-            pycodestyle ${package_path} || build_failed "pycodestyle failed in ${package_path}"
+            pycodestyle ${package_path} --max-line-length 120 || build_failed "pycodestyle failed in ${package_path}"
     fi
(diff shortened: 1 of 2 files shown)


march_ws/build.sh
```diff
@@ -11,33 +11,74 @@ build_failed () {
...
 # Run the tests, ensuring the path is set correctly.
 source devel/setup.bash || exit 1
 catkin_make run_tests && catkin_test_results || build_failed "Tests failed"
 
-for directory in src/*/launch
+# Run roslaunch check on all directories with a launch folder.
+for directory in $(find -O3 -L src/ -type d -name "launch")
 do
-    rosrun roslaunch roslaunch-check $directory || build_failed "roslaunch-check failed in directory $directory"
+    if [[ "$directory" != *"march_moveit"* ]]
```
Ishadijcks: Also create an ignore list for this check

Issue #168, opened 2019-03-01
Write Documention for all packages


TimBuckers:
- [ ] Write high level documentation for all packages
- [ ] Update current documentation to conform to doxygen style
- [ ] Review by other team member

Ishadijcks:
See https://github.com/project-march/tutorials/issues

PR #191, opened 2019-03-19
Feature/shared resources doxygen

TimBuckers:
General documentation about the shared resources.

Removed: `gait_enum.h` this file was not used. Otherwise the section about categories was not complete

connect https://github.com/project-march/march-iv/issues/168


march_ws/src/march_shared_resources/docs/pages/mainpage.dox
```diff
@@ -0,0 +1,11 @@
+/** @mainpage March Shared Resources
+ *
+ * The shared resources package consists of all resources that are used in multiple packages.
+ *
+ * The content of this package can be divided in a few categories:
+ * * <b>Custom ROS Messages</b>
+ * * <b>Custom ROS Services</b>
```
JitskedeVries: Maybe an example for a ROS Message and a ROS service

PR #210, opened 2019-05-15
PDB state and powernet msg

TimBuckers:
PDB message needed for the pdb state publisher


march_shared_resources/msg/PowerNet.msg
```diff
@@ -0,0 +1,7 @@
+string[] net_names
```
Ishadijcks: Why would a single net have multiple `net_names`?

PR #213, opened 2019-05-21
Add gait msgs to march_shared_resources

Ishadijcks:
Should be in the shared_resources instead of the gait_generator


march_shared_resources/msg/Setpoint.msg
```diff
@@ -0,0 +1,3 @@
+# A map of which joints have a user-defined setpoint at this time_from_start.
+duration time_from_start
+string[] joint_names
```
Ishadijcks:
```suggestion
string[] joint_names

```

PR #220, opened 2019-05-23
Add code coverage support for python tests

Ishadijcks:
Turns out the Testrunner class was not needed as it wasn't called at all.
`catkin_add_nosetests()` finds the tests without the need for a main.

I've also installed the `codecov` python module to upload reports, it's a bit cleaner.

nosetests generates quite a different report, so we use the `coverage` command to convert it to something codecov understands.

Also refactored python node. It now has a single executable script so the other files can stay non-executable and be found by nosetests coverage.


codecov.yaml
```diff
@@ -9,4 +9,5 @@ ignore:
   - "march_gait_scheduler/test/.*"
   - "march_gait_selection/test/.*"
   - "march_safety/test/.*"
+  - "**/__init__.py"
```
TimBuckers:
```suggestion
```

TimBuckers:
The code coverage report doesn't contain all files
image

PR #243, opened 2019-06-25
Add home gait action server

Ishadijcks:
Add a new topic to allow for homing gaits. The main difference is that they aren't transition validated against other subgaits


Ishadijcks:
Not needed, home gaits can be achieved from gait-files

PR #245, opened 2019-06-26
send alive message

TimBuckers:
Send alive message as developer ipd:
- Rate 20
- runs in thread
- always sending alive message

For the first training we only need developer ipd, later on we can make the alive message optional.


Ishadijcks:
Did you manage to fix the testing bug?

PR #259, opened 2019-07-03
Parse the urdf to check which temperature sensors we need to

### LINKED ISSUE #254: Temperature Safety subscribes to incorrect topics
Currently it expects a jointlist at the param `/sensors`. It should parse the urdf for all active joints instead.

Ishadijcks:
Closes #254 

Previously we checked the `/sensor` parameter which was only uploaded
by the fake temperature sensors

Add better sound files


march_safety/src/SafetyNode.cpp
```diff
@@ -29,7 +30,38 @@ int main(int argc, char** argv)
 
   std::vector<SafetyType> safety_list;
 
-  TemperatureSafety temperatureSafety = TemperatureSafety(&n, &safetyHandler);
+  int count = 0;
+  while (!n.hasParam("/robot_description"))
+  {
+    ros::Duration(0.5).sleep();
+    count++;
+    if (count == 10)
```
TimBuckers:
```suggestion
    if (count > 10)
```

PR #275, opened 2019-07-15
velocity limits increased, but lowered for the ankle

TimBuckers:
velocity limits:
- high and knee increased to 2.5
- ankle decreased to 0.6


TimBuckers:
hip_aa_velocity_limit are 1.0 do we want to change this value?

Ishadijcks:
Is this still relevant or can this be closed?

PR #277, opened 2019-07-16
Cache subgaits to massively improve response time

Ishadijcks:
Schedule offset can go from 0.9 to 0.2. At least on my machine, YMMV
image
:)

Please test this extensively, it is very possible this introduced some bugs that I can't think of in my current state.


Ishadijcks:
@bentebloks I think that error is fixed now, can you try again?

PR #282, opened 2019-07-18
add margin for time offset between arduino and master pc

TimBuckers:
I noticed a offset between the time on the arduino and on the master pc.
I also found some open issues on time sync: https://github.com/ros-drivers/rosserial/issues/392
- [x] For now I just added a small margin.
- [x] Increased the input_device_connection_timeout in the tests, to make it more stable.

### DIFF
--- a/march_safety/src/InputDeviceSafety.cpp
+++ b/march_safety/src/InputDeviceSafety.cpp
@@ -49,7 +49,7 @@ void InputDeviceSafety::update()
     }
     // Check if the alive msg is not timestamped with a future time.
     // This can happen when one node is using sim_tim and others aren't.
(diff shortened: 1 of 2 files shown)


march_safety/src/InputDeviceSafety.cpp
```diff
@@ -49,7 +49,8 @@ void InputDeviceSafety::update()
     }
     // Check if the alive msg is not timestamped with a future time.
     // This can happen when one node is using sim_tim and others aren't.
-    if (ros::Time::now() < time_last_alive)
+    // Add 1 second extra margin for stamp offset between board and PC
```
TimBuckers:
```suggestion
    // Add small margin to take the stamp offset between board and PC into account
```

PR #287, opened 2019-07-22
Automatically start rosbag node in launch file

### LINKED ISSUE #290: Automatically logging
Start a rosbag node in the marchiv launch file and the march_headless launch file.

JitskedeVries:
## Description

It starts a rosbag node when launching the exoskeleton. This node logs all topics and stores it in a default folder on the exoskeleton. The name of the default folder must be changed before starting the exoskeleton. But logging can never be forgotten anymore!

Closes #290


Ishadijcks:
The reason this wasn't added before is that it assumes something about the computer we're running on. What if I want to log the simulation locally and my computer is not named `march`? The bag will throw an error.

If you can resolve this dependency, it can be merged to develop.

sopje:
> The reason this wasn't added before is that it assumes something about the computer we're running on. What if I want to log the simulation locally and my computer is not named `march`? The bag will throw an error.
> 
> If you can resolve this dependency, it can be merged to develop.

Any ideas in how to do this?

JitskedeVries:
> > The reason this wasn't added before is that it assumes something about the computer we're running on. What if I want to log the simulation locally and my computer is not named `march`? The bag will throw an error.
> > If you can resolve this dependency, it can be merged to develop.
> 
> Any ideas in how to do this?

I have some ideas, but they are both not ideal. 
- In the launch file you could search for a package directory. We could dedicate a new or existing package to all the logs. Since the launch file then searches for a package directory, the path will be the same for all computers.
- You can also get the directoryname of the launch file and store the log files over there. But this is even uglier than storing it in a package.

I prefer the first one, but I think it is even better to store the logs somewhere outside of the workspace. I have googled this, but could not find a solution yet.

@sopje @Ishadijcks What are your thoughts on this?

Ishadijcks:
Great, if you could add it to the headless as well I can approve it.

Ishadijcks:
Nice, could you update the documentation to include a section where we explain where it logs when you don't specify it?

PR #288, opened 2019-07-23
Add message to display command after limits

JitskedeVries:
### Short discription
Add a new message that can be used in the hardware interface to display the commands that are really send towards the iMotionCube. 

### Explanation
First the trajectory is send towards the trajectory controller, this trajectory controller calculates dependently on the error which command must be sent towards the IMC. This calculated command can be seen in several topics. But before the command is written towards the IMC, the limits interface will check if this command is possible in combination with the current position, current speed and all the position, velocity and effort limits. This limit interface adjusts the command before it is send towards the IMC. This new feature makes is possible to see what command is actually sent after the limits.

### Hardware Interface
This message is used in the hardware interface on the feature/realtime-pub-afterlimit-command branch. In the ```void MarchHardwareInterface::write(ros::Duration elapsed_time)``` method in the march_hardware_interface.cpp: https://github.com/project-march/hardware-interface/pull/116


martijnvandermarel:
@JitskedeVries Can you link to the PR/branch in the hardware interface? Hard to review otherwise

march_shared_resources/msg/AfterLimitCommand.msg
```diff
@@ -0,0 +1,3 @@
+string[] name
+float64[] position_command
```
martijnvandermarel: Are the position_command and effort_commands really 64 bit (long floats) or are they normal floats (32-bit)?

Bart-van-Ingen: The definition of the commands can be found in the hardware interface and are as such:
```
  std::vector<double> joint_position_command_;
  std::vector<double> joint_velocity_command_;
```

Bart-van-Ingen:
> Message name is not descriptive enough

@Ishadijcks what needs to added then? would you consider AfterLimitControllerCommand sufficient?

Ishadijcks:
`AfterLimitJointCommand` would be better I think. This also implies what the command is send to.

But this whole feature is only to inspect the effects of the SafetyLimits right? I wonder if there is a cleaner way to fix this, like adjusting the desired values before they were published...

march_shared_resources/msg/AfterLimitJointCommand.msg
```diff
@@ -0,0 +1,3 @@
+string[] name
+float64[] position_command
+float64[] effort_command
```
Ishadijcks:
```suggestion
float64[] effort_command

```

PR #289, opened 2019-07-23
Adjustment of k_velocity and effort limit

JitskedeVries:
### Short description
Adjustments of the k_velocity and effort limit, since they are very important for torque mode.

### Effort limit
The effort limit is adjusted to 21840 IU, this converts to 20 A according to the IMC bible. MARCH 3 used 20 A as their nominal current setting for the HFE/KFE joints in EasyMotionStudio. Therefore we decided to raise the effort limit from 8000 IU to 21840 IU. 
Note: It is allowed to cross the nominal current for during some gaits. The peak current is about 40 A. Thus we could increase this limit when we find it necessary. This is not yet needed for airgaiting (and normal walking with Sjaan I think).

### K velocity
The k_velocity is adjusted to 20000. With a k_velocity of 5 the joint is unable to move, since the joint is never allowed to send an effort which is higher than 1800. 1800 IU is the amount of effort with which the joint starts moving. The k_velocity is approximately determined at 20000. This way the is able to move with it's maximum effort, but still the effort is limited when the joint is at it's maximum velocity.
Note: also this value could be determined with more reasoning, but I think it is not necessary for now.


march_description/urdf/test-joint.xacro
```diff
@@ -18,8 +18,8 @@
     <xacro:property name="joint_buffer" value="${3*pi/180}"/> <!-- rad -->
     <xacro:property name="joint_lower_limit" value="-0.5673"/> <!-- rad -->
     <xacro:property name="joint_upper_limit" value="0.3549"/> <!-- rad -->
-    <xacro:property name="joint_effort_limit" value="1000.0"/>
-    <xacro:property name="joint_velocity_limit" value="1.5"/> <!-- rad/s -->
+    <xacro:property name="joint_effort_limit" value="21840"/> <!-- IU = 20 A -->
+    <xacro:property name="joint_velocity_limit" value="0.5"/> <!-- rad/s -->
```
sopje:
```suggestion
    <xacro:property name="joint_velocity_limit" value="2.0"/> <!-- rad/s -->
```

PR #295, opened 2019-07-24
Gait selection node returns False if gait is not scheduled

### LINKED ISSUE #138: Action in state machine: aborted is interpreted as successful
## Describe the bug
The PoseState in the state machine package action call is not working properly. When the action returns aborted the state machine interprets this as successful. The bug can origin from the FeedbackActionState this is the parent class of PoseState. 

## To Reproduce
2. Remove a group state from march.srdf in the march_moveit package 
For example remove this:
`<group_state name="stand" group="full_exo">
        <joint name="left_ankle" value="-0.1439" />
        <joint name="left_hip" value="0" />
        <joint name="left_knee" value="0" />
        <joint name="right_ankle" value="0" />
        <joint name="right_hip" value="0" />
        <joint name="right_knee" value="0" />
    </group_state>` 
3. roslaunch march_launch march_iv.launch 
4. rosrun rqt_input_device rqt_input_device 
3. Navigate through the state machine with the input device. Until you execute the delete gait (stand).

## Expected behavior
The state machine will now **think** the gait is executed `successful`. However the gait is `aborted`. Therfore, the state machine should go to the `unknown` state.

## Additional context 
In the console a warning about the missing pose should appear, when trying to execute a gait which doesn't exist.

Ishadijcks:
Closes #138

Before the result of the scheduler was ignored, now it's handled properly

Related to https://github.com/project-march/state-machine/pull/69

### DIFF
--- a/march_gait_selection/src/march_gait_selection/GaitSelectionNode.py
+++ b/march_gait_selection/src/march_gait_selection/GaitSelectionNode.py
@@ -30,12 +30,16 @@ class PerformGaitAction(object):
         rospy.loginfo("Trying to schedule subgait %s/%s", goal.name, goal.subgait_name)
         if not self.gait_selection.validate_gait_by_name(goal.name):
             rospy.logerr("Gait %s is invalid", goal.name)
-            self.action_server.set_succeeded(False)
+            self.action_server.set_aborted("Gait " + str(goal.name) + "is invalid")
             return False
         subgait = self.gait_selection.get_subgait(goal.name, goal.subgait_name)
         trajectory_result = self.schedule_gait(goal.name, subgait)
-        # @TODO(Isha, Tim) monitor the scheduled gait and pass feedback to the state machine.
-        self.action_server.set_succeeded(True)
+        if trajectory_result:
+            self.action_server.set_succeeded(trajectory_result)
+            return True
+        else:
+            self.action_server.set_aborted(trajectory_result)
+            return False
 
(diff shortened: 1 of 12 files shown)


march_gait_selection/src/march_gait_selection/GaitSelectionNode.py
```diff
@@ -30,12 +30,16 @@ def target_gait_callback(self, goal):
...
-            self.action_server.set_succeeded(False)
+            self.action_server.set_aborted("Gait " + str(goal.name) + "is invalid")
             return False
         subgait = self.gait_selection.get_subgait(goal.name, goal.subgait_name)
         trajectory_result = self.schedule_gait(goal.name, subgait)
-        # @TODO(Isha, Tim) monitor the scheduled gait and pass feedback to the state machine.
-        self.action_server.set_succeeded(True)
+        if trajectory_result:
+            self.action_server.set_succeeded(trajectory_result)
+            return True
```
martijnvandermarel: Function didn't return anything before, now it does. Why? Is the returned value now used?

martijnvandermarel: Ah, I think I found this in the referenced PR. Was confused because I couldn't find it in this PR

Ishadijcks: It actually was, see line 34. Turns out it is not needed though, I've removed it

PR #297, opened 2019-07-24
Check state instead of result to prevent wrong abortion of scheduled gait

Ishadijcks:


march_gait_selection/src/march_gait_selection/GaitSelectionNode.py
```diff
@@ -32,11 +33,16 @@ def target_gait_callback(self, goal):
...
+        if trajectory_state == actionlib.GoalStatus.SUCCEEDED:
+            self.action_server.set_succeeded(trajectory_state)
         trajectory_result = self.schedule_gait(goal.name, subgait)
-        if trajectory_result:
+
+        if trajectory_result == GoalStatus.SUCCEEDED:
             self.action_server.set_succeeded(trajectory_result)
         else:
-            self.action_server.set_aborted(trajectory_result)
+            self.action_server.set_aborted(trajectory_state)
```
martijnvandermarel: Why do state here instead of result?

Ishadijcks: The result is empty as we don't calculate anything, the state determines if the action was completed or aborted.

march_gait_selection/launch/march_gait_selection.launch
```diff
@@ -7,7 +7,7 @@
     />
     <arg
             name="gait_file_directory"
-            default="airgait"
+            default="training"
```
Ishadijcks:
```suggestion
            default="airgait"
```

march_gait_selection/src/march_gait_selection/GaitSelectionNode.py
Ishadijcks: Ah I see what you mean, I've committed wrong, fixed!

march_description/urdf/march-iv.xacro
```diff
@@ -81,7 +81,7 @@
         <origin xyz="0 0 ${lower_leg_height+upper_leg_height+hover_height}" rpy="0 0 0"/>
     </joint>
 
-    <joint name="left_hip_aa" type="revolute">
+    <joint name="left_hip_aa" type="fixed">
```
Ishadijcks:
```suggestion
    <joint name="left_hip_aa" type="revolute">
```

Ishadijcks:
The struggles of trying to work during a training, can you rereview?

PR #303, opened 2019-07-25
Feature/trajectory controller tolerance safety

### LINKED ISSUE #217: Failure of the follow trajectory action is not handled
In this case we probably want to stop (perhaps throw error for the safety node)

Bart-van-Ingen:
## Issues
Closes #217

## Description
The following feature adds a safety feature that checks whether the trajectory controller is within it trajectory constraints. Once the constraints have been passed a position hold command is sent, completed and then the trajectory controller is stopped. Stopping the controller ensure that no new command can be sent afterwards. The controller will need to be started to take new commands.

## Features
- TrajectorySafety Class used by SafetyHandler in SafetyNode
  - Pulls trajectory constraints from controller definition from parameter server
  - compares constraint with error calculated by controller
  - If constraints passed, hold position command sent and trajectory controller stopped  

## Documentation
https://project-march.github.io/tutorials/doc/march_packages/march_safety.html


march_safety/src/SafetyHandler.cpp
```diff
@@ -57,3 +58,64 @@ void SafetyHandler::publishNonFatal(std::string message)
...
+
+void SafetyHandler::stopController(const std::string& stop_controller, float stop_trajectory_duration)
+{
+  // stopping the controller will execute one last update() call once. Depending on the interface either the position
+  // command or the effort command will afterwards stay constant.
+  ros::ServiceClient client = this->n->serviceClient<controller_manager_msgs::SwitchController>(
+      ros::this_node::getNamespace() + "/controller_manager/switch_controller");
+  controller_manager_msgs::SwitchController ctr = controller_manager_msgs::SwitchController();
+
+  if (getControllerStatus(stop_controller) == "stopped")
```
Ishadijcks: Is this a string instead of const/enum?

Bart-van-Ingen:
Rrbot is the robot used by ros_controllers to test the controllers. The reason why I am using it here is because it is simple and I am able to copy the hardware interface. It is a simple way of loading a trajectory controller. What would the benefit be of using our own robot? I am in essence only checking the response of the node to an arbitrary trajectory controller. The robot used is of no importance.

march_safety/src/SafetyHandler.cpp
Bart-van-Ingen: the response of the controller manager indeed returns a string and not an enum. This is therefore a design detail of the controller manager and not per se of the safety node. Are you suggesting that a different string comparison should be used?

Ishadijcks:
Can you explain why we need to stop the controller? Why is holding the trajectory not enough?

march_safety/src/SafetyNode.cpp
```diff
@@ -22,14 +24,20 @@ int main(int argc, char** argv)
...
+  std::string controller_name;
 
   // Create an error publisher to notify the system (state machine) if something is wrong
   ros::Publisher error_publisher = n.advertise<march_shared_resources::Error>("/march/error", 1000);
   ros::Publisher sound_publisher = n.advertise<march_shared_resources::Sound>("/march/sound/schedule", 1000);
   ros::Publisher gait_instruction_publisher =
       n.advertise<march_shared_resources::GaitInstruction>("/march/input_device/instruction", 1000);
 
-  SafetyHandler safetyHandler = SafetyHandler(&n, &error_publisher, &sound_publisher, &gait_instruction_publisher);
+  n.getParam(ros::this_node::getName() + std::string("/controller_name"), controller_name);
```
Ishadijcks: I don't see where the param is uploaded.
Also what if it can't find our controller, I believe that should be a fatal.

march_safety/src/TrajectorySafety.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+  bool stop_trajectory_duration_status =
+      n->getParam(controller_path + "/stop_trajectory_duration", this->stop_trajectory_duration);
+
+  this->trajectory_subscriber = n->subscribe<control_msgs::JointTrajectoryControllerState>(
+      controller_path + "/state", 1000, &TrajectorySafety::trajectoryCallback, this);
+
+  if (stop_trajectory_duration < 0)
+  {
+    ROS_WARN("stop_trajectory_duration has been set below zero ");
```
Ishadijcks: Maybe set it to 0?

march_safety/src/TrajectorySafety.cpp
```diff
@@ -0,0 +1,70 @@
...
+  for (auto it = this->joint_names.begin(); it != this->joint_names.end(); it++)
+  {
+    double joint_tolerance;
+    if (n->getParam(controller_path + "/constraints/" + *it + "/trajectory", joint_tolerance))
+    {
+      this->trajectory_tolerances[*it] = joint_tolerance;
+    }
+    else
+      ROS_WARN("tolerance of joint %s cannot be found. Make sure it is indicated in the controller yaml",
+               getJointName(it).c_str());
```
Ishadijcks: Should this be a fatal?

march_safety/src/SafetyNode.cpp
Bart-van-Ingen: the param is currently uploaded through the launch file. The fatal is a good idea!

Bart-van-Ingen:
> Can you explain why we need to stop the controller? Why is holding the trajectory not enough?
in my opinion the controller needs to be stopped as to ensure that any new trajectories that are sent to it are not performed. If a new trajectory (for example a subgait) is sent after the hold command, without the controller being stopped, the controller will react to this new command and try to complete it. I see it as an extra safety measure as to ensure that once the tolerances have not been met that the device will definitely not move. 
In order to move again a new method will have to be written where the controller is started and thus where it can accept new trajectory commands. However I do not believe this is within the scope of this PR.

Ishadijcks:
Note that the state machine will go to the unknown state as an error is thrown. The only way to then send trajectories is through the action topic. This is always possible anyway, so that wouldn't become a greater risk once we have decided to stop moving, except for some bugs.

If this is implemented with stopping, we would need to design the state machine to be able to react to this change. Get out of the healthy state into some sort of unhealthy, where a developer could then send a command to restart the controller which would bring it to unknown. Do you agree with that?

march_safety/src/TrajectorySafety.cpp
Bart-van-Ingen: this should maybe become a fatal, since a -1 number will cause the trajectory controller to crash. 
It would also be possible to set the parameter to 0 on the parameter server in order to prevent the crash, however this will lead to changing of parameters specified by the user, which i do not think is preferable.

Bart-van-Ingen:
> Note that the state machine will go to the unknown state as an error is thrown.

Is this an error thrown by the safety node or by the trajectory controller?

> The only way to then send trajectories is through the action topic. This is always possible anyway, so that wouldn't become a greater risk once we have decided to stop moving, except for some bugs.

In the case that the controller is stopped, a client server using the action topic is still able to be active, however it is not possible to send commands. Sending commands will therefore not have any effect.

> Get out of the healthy state into some sort of unhealthy, where a developer could then send a command to restart the controller which would bring it to unknown. Do you agree with that?

This seems reasonable. it does however depend on the answer to my first question. If the safety node is throwing all errors on which the state machine reacts, code will have to be add to the safety node in which it will have to induce a transition to unknown state. 
Would the restarting of the controller take place in the march_safety node?

Ishadijcks:
> Is this an error thrown by the safety node or by the trajectory controller?

Safety node

> In the case that the controller is stopped, a client server using the action topic is still able to be active, however it is not possible to send commands. Sending commands will therefore not have any effect.

Sorry, I meant the command topic. I worded it poorly as I meant it was possible during healthy behaviour.


Based on the type of error the SafetyNode throws, the SafetyState can send the statemachine to a specific state to wait on the command to restart the controller

Bart-van-Ingen:
> Based on the type of error the SafetyNode throws, the SafetyState can send the statemachine to a specific state to wait on the command to restart the controller

yes this sounds good. Would the restarting of the controller take place in the march_safety node? Or will that need to be implemented by the state_machine?

Ishadijcks:
I think the state machine, the safety node has handled his responsibility at this moment. The rest of the behaviour (listening to a reset command, performing the restart) would be more suited for the state machine.

Bart-van-Ingen:
I have implemented the message to the state machine. It currently throws a non fatal error and indicates in the message that the controller, with its specific name, has been stopped. Is this sufficient for the state machine to catch and process?

march_safety/src/TrajectorySafety.cpp
```diff
@@ -17,7 +17,8 @@ TrajectorySafety::TrajectorySafety(ros::NodeHandle* n, SafetyHandler* safety_han
   controller_client = new ControllerClient(controller_path + "/follow_joint_trajectory", true);
 
   // wait for action server to come up, if so the controller is available
-  if (!controller_client->waitForServer(ros::Duration(0.5)))
+  ros::Duration(0.5).sleep();
+  if (!controller_client->isServerConnected())
```
Ishadijcks: Why was this changed?

Bart-van-Ingen: i was seeing if there was a timing issue with Travis that I wasn't having with my local build. In the end I opted for a different method that i find cleaner than what is presented here. It also allowed for a successful build!

march_safety/src/TrajectorySafety.cpp
```diff
@@ -0,0 +1,81 @@
+// Copyright 2019 Project March.
+
+#include <control_msgs/JointTrajectoryControllerState.h>
```
TimBuckers:
```suggestion
```
This include is already defined in the TrajectorySafety.h

march_safety/src/TrajectorySafety.cpp
```diff
@@ -0,0 +1,81 @@
...
+  : safety_handler(safety_handler), joint_names(std::move(joint_names))
+{
+  n->getParam(ros::this_node::getName() + std::string("/controller_name"), this->controller_name);
+  controller_path = "/" + controller_name;
+
+  bool stop_trajectory_duration_status =
+      n->getParam(controller_path + "/stop_trajectory_duration", this->stop_trajectory_duration);
+
+//  // check whether the controller is available, through the use of an action client
+//  controller_client = new ControllerClient(controller_path + "/follow_joint_trajectory", true);
```
TimBuckers: why is there commented code?

march_safety/src/TrajectorySafety.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+TrajectorySafety::TrajectorySafety(ros::NodeHandle* n, SafetyHandler* safety_handler,
+                                   std::vector<std::string> joint_names)
+  : safety_handler(safety_handler), joint_names(std::move(joint_names))
+{
+  n->getParam(ros::this_node::getName() + std::string("/controller_name"), this->controller_name);
+  controller_path = "/" + controller_name;
+
+  bool stop_trajectory_duration_status =
+      n->getParam(controller_path + "/stop_trajectory_duration", this->stop_trajectory_duration);
```
TimBuckers: /stop_trajectory_duration is a boolean? duration doesn't sound like a boolean.

march_safety/src/TrajectorySafety.cpp
```diff
@@ -0,0 +1,81 @@
...
+    else
+      ROS_FATAL("tolerance of joint %s cannot be found. Make sure it is indicated in the controller yaml",
+                getJointName(it).c_str());
+  }
+}
+
+void TrajectorySafety::trajectoryCallback(const control_msgs::JointTrajectoryControllerStateConstPtr& msg)
+{
+  int counter = 0;
+  for (auto it = msg->joint_names.begin(); it != msg->joint_names.end(); ++it)
```
TimBuckers: Personally I find this code not easy to read. 
(I see you assume that the index of the error.position and the joint name are equal, right?)
You could do something like this:
```void TrajectorySafety::trajectoryCallback(const control_msgs::JointTrajectoryControllerStateConstPtr& msg)
{
  for(int i = 0; i < joint_names.size(); i++){
    std::string joint_name = joint_names[i];
    this->position_errors[joint_name] = msg->error.positions[i];
  }
  toleranceCheck();
}

File: march_shared_resources/include/march_shared_resources/TopicNames.h [hunk 10]
```diff
@@ -48,6 +48,7 @@ const char* const gait_denied = "march/gait/denied";
 const char* const follow_joint_trajectory_execution = "/march/trajectory_controller/follow_joint_trajectory/goal";
 const char* const follow_joint_trajectory_execution_states = "/march/trajectory_controller/follow_joint_trajectory/"
                                                              "feedback";
+const char* const trajectory_controller = "/march/trajectory_controller";
```
TimBuckers:
```suggestion
```
This const is not used. Also we are not really using TopicNames anymore

march_safety/src/TrajectorySafety.cpp
Bart-van-Ingen: maybe its a little confusing but the boolean is for the stop_trajectory_duration_**status** . The `getParam` method returns a boolean on whether it was able to get the parameter from the parameter server. Depending on if this was possible either an error is thrown or the code continues.

march_safety/src/SafetyNode.cpp
```diff
@@ -22,14 +24,20 @@ int main(int argc, char** argv)
...
+  ros::Publisher stop_trajectory_publisher =
```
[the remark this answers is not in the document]
Bart-van-Ingen: good point! changed it.

march_safety/src/TrajectorySafety.cpp
Bart-van-Ingen: removed

march_safety/src/TrajectorySafety.cpp
Bart-van-Ingen: removed

march_safety/src/TrajectorySafety.cpp
TimBuckers: Makes sense

march_safety/CMakeLists.txt
```diff
@@ -92,5 +109,4 @@ if (CATKIN_ENABLE_TESTING)
     endif()
 
 
-endif ()
-
+endif ()
```
Ishadijcks:
```suggestion
endif ()

```

march_safety/include/march_safety/TrajectorySafety.h
```diff
@@ -0,0 +1,39 @@
+// Copyright 2018 Project March.
```
Ishadijcks:
```suggestion
// Copyright 2019 Project March.
```

Ishadijcks:
Fix the merge conflict please

march_safety/include/march_safety/TrajectorySafety.h
```diff
@@ -0,0 +1,39 @@
...
+#define SRC_TRAJECTORYSAFETY_H
+
+#include "ros/ros.h"
+#include "SafetyType.h"
+#include "SafetyHandler.h"
+#include "control_msgs/JointTrajectoryControllerState.h"
+#include <actionlib/client/simple_action_client.h>
+#include <control_msgs/FollowJointTrajectoryAction.h>
+
+#include <march_shared_resources/TopicNames.h>
```
Ishadijcks:
```suggestion
```

march_safety/src/TrajectorySafety.cpp
```diff
@@ -0,0 +1,84 @@
...
+    ROS_FATAL("either the controller is unavailable, or the controller path is incorrect");
+    std::runtime_error("controller safety is unavailable, due to controller unavailable or wrong controller path");
+  }
+
+  this->trajectory_subscriber = n->subscribe<control_msgs::JointTrajectoryControllerState>(
+      controller_path + "/state", 1000, &TrajectorySafety::trajectoryCallback, this);
+
+  if (this->stop_trajectory_duration < 0)
+  {
+    ROS_FATAL("stop_trajectory_duration has been set below zero ");
```
Ishadijcks: Throw an error here if it's a fatal?

march_shared_resources/include/march_shared_resources/TopicNames.h
```diff
@@ -77,7 +76,6 @@ const char* const request_gait_file = "march/request_gait_file";
 namespace ActionNames
 {
 const char* const schedule_gait = "march/gait/schedule";
-
 };
```
Ishadijcks: Revert these changes to resolve the merge conflict

march_safety/src/TrajectorySafety.cpp
```diff
@@ -0,0 +1,84 @@
...
+  for (auto it = this->joint_names.begin(); it != this->joint_names.end(); it++)
+  {
+    double joint_tolerance;
+    if (n->getParam(controller_path + "/constraints/" + *it + "/trajectory", joint_tolerance))
+    {
+      this->trajectory_tolerances[*it] = joint_tolerance;
+    }
+    else
+      ROS_FATAL("tolerance of joint %s cannot be found. Make sure it is indicated in the controller yaml",
+                getJointName(it).c_str());
```
Ishadijcks: Always use brackets for control flow.

Fatal without crashing

TimBuckers:
Not needed anymore due to: https://github.com/project-march/march-iv/pull/340

PR #309, opened 2019-08-02
Add git branch check

### LINKED ISSUE #300: More software checks
I don't know if all these things are possible to check, but here are some suggestions:
- Check on which branches the repos are on, must be either on develop or on the most recent training branch
- Check joint inside softlimits before launching the hardware interface (could save time).
- Check if the logfiles are stored on an existing location. Thus if rosbagnode which is launched in the launch file is working.

Ishadijcks:
Closes #300 

Note that this check is run on the exoskeleton


Ishadijcks:
git-branch-check

Issue #311, opened 2019-08-05
Gait 'succeeded' when not actually succeeded due to controller joint mismatch


martijnvandermarel:
When a mismatch is present between the joints defined in controllers.yaml and those in the urdf and a gait is played (in this case homing stand), the gait 'succeeds' and the state machine transitions (in our case to standing) even though the exoskeleton hasn't physically moved

![Screenshot from 2019-08-05 15-38-22](https://user-images.githubusercontent.com/23376251/62468985-93418480-b797-11e9-8f68-b4aee01ae600.png)

bag file:
[_2019-08-05-15-36-55.zip](https://github.com/project-march/march-iv/files/3467827/_2019-08-05-15-36-55.zip)

Ishadijcks:
Did this occur naturally or did you force this with manual publishing somehow?

martijnvandermarel:
Normal operation (apart from the controller mismatch of course)

Ishadijcks:
Running with all joints? What did you expect to happend?

martijnvandermarel:
Yes, 8 joints in urdf, 6 in controllers.yaml. Expected would be imo either a fatal error or a 'gait failed' message with the state machine staying in unknown state. It is unsafe to continue operation when the state machine is in a different state than the actual exoskeleton

TimBuckers:
https://github.com/project-march/march-iv/pull/314 @martijnvandermarel

PR #314, opened 2019-08-05
Handle the gait action result correctly

TimBuckers:
Closes https://github.com/project-march/march-iv/issues/311


martijnvandermarel:
Looks good, is this tested on the simulation, and if so is the simulation representative for whether it would work on the exoskeleton?

Ishadijcks:
In theory they run the same controllers, can't hurt to check though.

TimBuckers:
https://github.com/project-march/march-iv/pull/325

PR #318, opened 2019-08-08
disable smach viewer in state machine when running headless

TimBuckers:
to make this works this PR is needed https://github.com/project-march/state-machine/pull/73

### DIFF
--- a/march_launch/launch/march_headless.launch
+++ b/march_launch/launch/march_headless.launch
@@ -15,5 +15,7 @@
             args="-a"
     />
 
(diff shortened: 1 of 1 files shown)


Ishadijcks:
I think more work is needed, like not starting the introspection server from the state machine

Ishadijcks:
http://wiki.ros.org/smach/Tutorials/Smach%20Viewer

TimBuckers:
see: https://github.com/project-march/state-machine/pull/73

PR #324, opened 2019-08-19
fix temperature above 2000 bug from the slaves

TimBuckers:
The temperature slaves sometimes send wrong values (first bit flips). These values are always above 2000 and because values above 2000 never occur we can filter them.

### DIFF
--- a/march_safety/src/TemperatureSafety.cpp
+++ b/march_safety/src/TemperatureSafety.cpp
@@ -36,7 +36,7 @@ void TemperatureSafety::temperatureCallback(const sensor_msgs::TemperatureConstP
 
   std::string error_message = getErrorMessage(temperature, sensor_name);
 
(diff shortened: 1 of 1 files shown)


march_safety/src/TemperatureSafety.cpp
```diff
@@ -36,6 +36,13 @@ void TemperatureSafety::temperatureCallback(const sensor_msgs::TemperatureConstP
 
   std::string error_message = getErrorMessage(temperature, sensor_name);
 
+  // TODO(Olaf) this is a temporary fix, this should be fixed locally on the slaves ask Electro.
```
Ishadijcks:
```suggestion
  // TODO(Olav) this is a temporary fix, this should be fixed locally on the slaves ask Electro.
```

Ishadijcks:
Can you update the description of this pr to explain why this is needed?

PR #326, opened 2019-08-19
crash if fatal

TimBuckers:


TimBuckers:
Crashing the node on fatal level, makes test fail. However, I couldn't find a way to catch this behaviour properly in a test. I think we should close this branch and feature? Not feasible for this year

Ishadijcks:
Difficult to say, what is worth more? The tests or this feature?

PR #328, opened 2019-08-21
increase hip and ankle temperature warning thresholds with 5 degrees

TimBuckers:
https://confluence.projectmarch.nl:8443/display/41TECH/Safety+Decisions


TimBuckers:
Only one response is executed, if warning and non-fatal have the same threshold only non-fatal response occurs. always in order of highest severity first.

Ishadijcks:
Makes sense!

PR #335, opened 2019-08-28
RQt training log

Olavhaasie:
## Description
Added an rqt plugin called `march_rqt_training_log` that has a text entry field and a `Log` button. Any text entered in the field will be published when the button is pressed to `/march/training/log`. This is useful to make notes during the training. For example when a soft limit is reached you could enter: _Sjaan's left foot got stuck on the third stairstep_. After the training you can find this message in the log files.

Currently the plugin is not launched during march launch, but it can be manually run with:

    rqt --standalone march_rqt_training_log

## Features

- New `march_rqt_training_log` package
- New rqt plugin for entering a log message and publishing it at `/march/training/log`


march_rqt_training_log/package.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>march_rqt_training_log</name>
+  <version>0.0.0</version>
+  <description>The march_rqt_training_log package</description>
+
+  <maintainer email="software@projectmarch.nl">Project March</maintainer>
```
Ishadijcks: You can add yourself as author and maintainer

march_rqt_training_log/package.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>march_rqt_training_log</name>
+  <version>0.0.0</version>
+  <description>The march_rqt_training_log package</description>
```
Ishadijcks: Update description

Olavhaasie:
Good ideas, I will look into this :+1:

PR #350, opened 2019-09-07
Improve the effort limit

JitskedeVries:
Increase the rotary and the linear effort limits to respectively 25 A and 28 A.
The IU are calculated with the technosoft formula.

image

The peak current of the drive is 40 A.

image

The ratio between IU and A is validated with a torque mode log file for both the linear and the rotary joints.


Olavhaasie:
@JitskedeVries is this still relevant?

PR #359, opened 2019-09-26
Correct masses for each of the parts.

Roelemans:
Closes PM-8 (https://jira.projectmarch.nl/browse/PM-8) and PM-7 (https://jira.projectmarch.nl/browse/PM-7). 

A mass parameter is now properly defined for each of the parts. The mass of the parts have been filled in according to https://confluence.projectmarch.nl:8443/display/41TECH/Weights+of+the+Exoskeleton.

Currently the simulation is faulty because the control is to weak for the new masses. Will be resolved in the next sprint.


Ishadijcks:
Since some masses are sums of other masses, it might be handy to define the original masses as variables and add them together. Will make it easier to change the masses later without additional calculations.

Roelemans:
@Ishadijcks I considered that, but in some cases it is more a clarification of what parts were included than an actual sum of masses, because these parts were weighed together. It all kinda depends on what parts you place on the scale together, so I figured I would just use the full weight instead of dividing it up in parts.

RutgerVanBeek:
I think you should either add to the confluence page how the masses are added (what you consider to be part of the lower leg for instance), with a sum weight, or you should do this in the code. Currently the masses in the URDF do not directly correspond to a value on the confluence page.

Ishadijcks:
You could also mark the grouping of weights visually on the urdf like this :p
image

Roelemans:
@RutgerVanBeek the sums have been added to confluence https://confluence.projectmarch.nl:8443/display/41TECH/Weights+of+the+Exoskeleton

PR #365, opened 2019-10-03
Fix/pm 63 correct inertias different axis

Roelemans:
Closes PM-63

## Description

Fix inertias for URDF coordinate system.


Roelemans:
While the moment of inertia around any axis cannot be negative, the intertia matrix of a point can have negative entries outside of its diagonal. So as long as Ixx, Iyy and Izz are non-negative we are not living in a black-white hole collision.

Ishadijcks:
interesting :)

Ishadijcks:
Also lucky us

PR #369, opened 2019-10-14
Small fixes from training

Olavhaasie:
## Description
This PR introduces some fixes I found during the fake training on Oktober 9.

## Changes
* Change default gait directory to `training`, that way we won't forget to set it and always use the safest and most tested gaits.
* Change workspace path on exoskeleton to home directory
* Fix `march_rqt_software_check` launch


march_launch/launch/march_headless.launch
```diff
@@ -26,10 +27,10 @@
             name="record"
             type="record"
             output="screen"
-            args="-a"
+            args="-a -o $(bag_prefix)"
```
Ishadijcks: Don't you want to do the same for `march_iv.launch`?

PR #377, opened 2019-10-26
Fix/pm 102 paused start option

RutgerVanBeek:
Closes PM-102

## Description

A checkbox is added to the launch menu, so one can choose whether the simulation should be started paused or start directly. 
## Changes

In the launch menu user interface file, the checkbox button has been added. 
In the launch_sm file the result of the checkbox is given as an argument to the simulation launch file.


Olavhaasie:
The simulation is not starting paused for me :crying_cat_face: also when I launch the `march_simulation.launch`, which should pause by default, which was introduced in project-march/simulation#25. So that's weird

Roelemans:
Did you do colcon build @Olavhaasie?

Olavhaasie:
Closed, since project-march/simulation#26 was chosen

PR #379, opened 2019-10-29
Add custom message for step height and length.

Roelemans:
Related to PM-49

## Description

Add a message containing step height and length. This message is used in the gait generator. 

## Changes

* Added the message FeetDistances.msg

### DIFF
--- a/march_shared_resources/CMakeLists.txt
+++ b/march_shared_resources/CMakeLists.txt
@@ -12,8 +12,8 @@ find_package(catkin REQUIRED COMPONENTS
 add_message_files(
     FILES
     AfterLimitJointCommand.msg
-    FeetDistances.msg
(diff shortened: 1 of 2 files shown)


Olavhaasie:
Why does this have to be a message?

Ishadijcks:
Is the gait generator finally being split into a client-server model like the gait selection? :heart_eyes:

Roelemans:
@Olavhaasie because a seperate node reads out tf frames to calculate the distances between the feet. This then has to be sent to the gait_generator node, I use this message for that purpose. If you have a suggestion for an alternative implementation, please let me know :)

Roelemans:
No longer necessary with new implementation in https://github.com/project-march/gait-generation/pull/69

PR #380, opened 2019-10-31
Fix: node initialization

Olavhaasie:
Partly closes PM-95

## Description
I rearranged the order of some action servers and topics subscriptions in order to make it easier to check when a node has started. This is important when all the nodes will be launched from the launch file instead of from the state machine. This way the state machine can check if the topics that are supposed to be available are available, so then it can transition to `HEALTHY`.

Furthermore, I downgraded some **info** logs to **debug** logs, since those were not necessary and only cluttering up the info level.

## Changes
* Rearrange safety node subscriptions
* Make `march_gait_scheduler` and `march_gait_selection` wait indefinitely for the action the server to be available, since there is no reason to continue if those are not available and many times they would continue and throw an error.
* Add `-q` parameter to `rosbag` to silence the `subscribed to ...` messages
* Remove the publisher in `march_gait_scheduler` to `/<controller_name>/command`, since the controller name parameter was not a valid one, so this did absolutely nothing.
* Add `delete` for allocated memory in `march_gait_scheduler`

_**Leave your code better than you found it.**_

### DIFF
--- a/march_gait_scheduler/src/GaitSchedulerNode.cpp
+++ b/march_gait_scheduler/src/GaitSchedulerNode.cpp
@@ -110,15 +110,20 @@ int main(int argc, char** argv)
   ros::init(argc, argv, "gait_scheduler_node");
   ros::NodeHandle n;
 
+  const std::string follow_joint_trajectory_topic = "/march/controller/trajectory/follow_joint_trajectory";
+
   scheduler = new Scheduler();
 
   scheduleGaitActionServer = new ScheduleGaitActionServer(n, "march/gait/schedule", &scheduleGaitCallback, false);
 
   followJointTrajectoryAction = new actionlib::SimpleActionClient<control_msgs::FollowJointTrajectoryAction>(
-      "/march/controller/trajectory/follow_joint_trajectory", true);
+      follow_joint_trajectory_topic, true);
 
   ROS_DEBUG("Wait on joint trajectory action server");
-  followJointTrajectoryAction->waitForServer();
+  while (ros::ok() && !followJointTrajectoryAction->waitForServer(ros::Duration(5.0)))
+  {
+    ROS_INFO_STREAM("Waiting for " << follow_joint_trajectory_topic << " to come up");
+  }
   ROS_DEBUG("Connected to joint trajectory action server");
 
   dynamic_reconfigure::Server<march_gait_scheduler::SchedulerConfig> server;
--- a/march_gait_scheduler/src/GaitSchedulerNode.cpp
+++ b/march_gait_scheduler/src/GaitSchedulerNode.cpp
@@ -1,6 +1,6 @@
 // Copyright 2018 Project March.
 
-#include "ros/ros.h"
+#include <ros/ros.h>
 #include <control_msgs/FollowJointTrajectoryAction.h>
 #include <ros/package.h>
 #include <std_msgs/Float64.h>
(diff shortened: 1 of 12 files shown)


march_safety/src/SafetyNode.cpp
```diff
@@ -46,8 +38,16 @@ int main(int argc, char** argv)
   std::vector<std::string> joint_names;
   n.getParam("/march/joint_names", joint_names);
```
Bart-van-Ingen: in the gait scheduler node you are waiting indefinitely for the server to start, while here you are giving the safety node 5 second to gather the joint names from the parameter server. Why the difference between the two?

march_rqt_launch_menu/src/march_rqt_launch_menu/launch_menu.py
```diff
@@ -100,7 +100,7 @@ def get_run_configuration(self):
 
     @staticmethod
     def publish_to_parameter_server(dictionary):
-        rospy.loginfo("Uploading the following launch_settings to the parameter server: " + str(dictionary))
+        rospy.logdebug("Uploading the following launch_settings to the parameter server: " + str(dictionary))
```
RutgerVanBeek: Personally I think this is nice to have as an info message.

march_gait_scheduler/src/GaitSchedulerNode.cpp
```diff
@@ -9,16 +9,15 @@
...
 #include <dynamic_reconfigure/server.h>
 
-#include <march_shared_resources/GaitAction.h>
-#include <march_shared_resources/GaitGoal.h>
-#include <march_gait_scheduler/Scheduler.h>
-#include <march_gait_scheduler/SchedulerConfig.h>
+#include "march_shared_resources/GaitAction.h"
+#include "march_shared_resources/GaitGoal.h"
+#include "march_gait_scheduler/Scheduler.h"
+#include "march_gait_scheduler/SchedulerConfig.h"
```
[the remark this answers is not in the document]
Olavhaasie: In C++ you can `#include` in two ways, with `<>` or with `""`, the first will look for system and installed header files, while the second will start searching locally. They achieve the same eventually, however this is more convention. I see that I edited it here while it still happens on a lot of places, but

_**Leave your code better than you found it.**_

march_safety/src/SafetyNode.cpp
Olavhaasie: Good question, the safety node should also wait indefinitely for this parameter, since it cannot continue without. That is why it throws `std::runtime_error`, however that would be flaky, when the hardware takes longer to launch.

march_rqt_launch_menu/src/march_rqt_launch_menu/launch_menu.py
Olavhaasie: @Roelemans what do you think?

march_safety/src/SafetyNode.cpp
Bart-van-Ingen: I agree with the flakyness(?), this is better!. should a similar while loop be considered in the gait scheduler? so the user can see whether it is stuck with reaching the server or not?

march_rqt_launch_menu/src/march_rqt_launch_menu/launch_menu.py
Roelemans: Nah if you want to know what options you selected, you can look back at the debug log

march_safety/src/SafetyNode.cpp
Olavhaasie: Good point, I have now implemented a similar loop in `gait_scheduler` and `gait_selection`

PR #382, opened 2019-11-05
Feature: gain scheduling

Deakensbubbles:
Closes PM-

## Description

A primitive form of gainscheduling able to change the P, I and D values of the controller based on the selected gait. at the moment the corresponding gain values to each gait are arbitrarily chosen. The " lookup table" used to chose these values is hardcoded in the node and will be replaced with a config file asap.

## Changes

A ROS package containing a Python package with a ROS Node that changes the PID values based on the selected gait.

PR #383, opened 2019-11-05
Add installing python-lxml

Olavhaasie:
## Description

For some weird reason the script wouldn't install the `python-lxml` dependency, which is why the build failed in #382. This issue is not caused by that PR, but I think by a change in the `urdf-parser-py` package from [this commit](https://github.com/ros/urdf_parser_py/commit/87b2913d5b5cc06178569191e33b5fded2675b13#diff-34ae68f4adad56c25c5bc05dcb64794eR28). :thinking: 

Anyway I found a quickfix.
## Changes

* Added manual install of `python-lxml` package

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ git:
 env:
   global:
     - ROS_REPO=ros
+    - ADDITIONAL_DEBS=python-lxml
     - CCACHE_DIR=$HOME/.ccache # enables C/C++ caching in industrial_ci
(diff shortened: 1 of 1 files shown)


Deakensbubbles:
Would this require any and all packages we build to depend on the python-lxml as well? Or was this dependency only required in Travis?

Olavhaasie:
This would be required for packages that use `urdf_parser_py` dependency

PR #385, opened 2019-11-08
Fix: more explainatory invalid transition error

RutgerVanBeek:
Helps with PM-122

## Description

When a gait is selected in the RQT gait selector, it is checked it its valid. One of these checks is whether the transition between subgaits is valid. This checks, whether the position, velocity, acceleration and effort of all the joints at the end of the old subgait is equal to the start of the new one. When there is an inequality just that error is printed. This PR makes sure the exact mistake is printed, so that it can be fixed more easily. 
## Changes

In the Transition test in the gait selection node, extra info messages are added when the test fails. 

This can be tested by using the branch RT-high-step for the gait-files where there is a wrong gait currently.


march_gait_selection/src/march_gait_selection/GaitSelection.py
```diff
@@ -278,7 +278,48 @@ def validate_trajectory_transition(old_trajectory, new_trajectory):
...
+        if (last_old_point_positions == first_new_point_positions and
+                last_old_point_velocities == first_new_point_velocities and
+                last_old_point_accelerations == first_new_point_accelerations and
+                last_old_point_effort == first_new_point_effort):
+
+            return True
+
+        else:
+            if last_old_point_positions != first_new_point_positions:
+                rospy.loginfo("Positions end old subgait do not match start new subgait: ")
```
Olavhaasie: Shouldn't these be `rospy.logwarn` or `rospy.logerr`? Since something fails. I would also recommend logging all this information in one log statement, otherwise it can be hard to find in the logs when other stuff is printed inbetween.

RutgerVanBeek: > Shouldn't these be `rospy.logwarn` or `rospy.logerr`? Since something fails. I would also recommend logging all this information in one log statement, otherwise it can be hard to find in the logs when other stuff is printed inbetween.

There is already an error thrown, these messages only add additional information to the error message. I can make these error messages as well, but I don't see the benefit.

Olavhaasie: What do you think about returning the complete error message from this function and then printing it when it failed on line 259? Then it will be one error message with complete explanation, which is much easier to read and debug.

Roelemans:
@RutgerVanBeek veel plezier met de falende build fixen.

march_gait_selection/launch/march_gait_selection.launch
```diff
@@ -7,7 +7,7 @@
     />
     <arg
             name="gait_file_directory"
-            default="training"
+            default="test_versions-v"
```
Roelemans: This is a mistake I think?

PR #390, opened 2019-11-11
Remove deprecated messages

Olavhaasie:
Closes PM-82

## Description
Removed deprecated and unused messages

## Changes
* Removed messages, services and actions that were no longer used
* Also from the `CMakeLists.txt`


Olavhaasie:
> If the messages are redundant they should indeed be removed. However, some seem interesting for future. were they actively used in the past or just one way feedback for developers?

I am not sure when they were used and for what purposes, you could ask @Ishadijcks or @TimBuckers.

PR #392, opened 2019-11-12
Add gait_type to subgait message.

Roelemans:
Necessary for https://github.com/project-march/gait-generation/pull/72

## Description

Add gait_type to a subgait message. The gait type will be used for gain scheduling in the future.

## Changes

* Add gait type to the subgait message.


Ishadijcks:
Is this a string or really an enumeration?

PR #396, opened 2019-11-18
Added paused argument

RutgerVanBeek:
Fix

## Description

When the simulation is started without the gazebo UI it is easier if the simulation time is already running and it doesn't start paused. Therefore the argument of whether the simulation should start paused should be equal to whether there is a gazebo UI. 
## Changes

Made the argument of paused equal to whether there is a gazebo UI.


Roelemans:
Olav is wise.

PR #398, opened 2019-11-18
Center of mass calculation

RutgerVanBeek:
Closes PM-115

## Description

Adds a class, CoMCalculator, which can calculate the center of mass based on the positions of links from the tf transforms and the masses from the URDF. A marker is created and published to march/com_marker. Rviz can use this to plot a marker add the center of mass. For the marker to be shown in Rviz it should be added like a robotmodel and the topic should be set to march/com_marker. From the marker message the position of the CoM can also be extracted, such that the movement of the CoM can be analysed during gaits. Next to that the CoM can be used in balance calculations. 
## Changes

The data collector node is refactored to a class. 
The ComCalculator class is added. 
Package dependencies have been added.


march_data_collector/src/march_data_collector/CoMCalculator.py
```diff
@@ -0,0 +1,72 @@
...
+import geometry_msgs.msg
+import tf2_ros
+import tf2_geometry_msgs as tf_geo
+from urdf_parser_py.urdf import URDF
+from visualization_msgs.msg import Marker
+
+
+class CoMCalculator(object):
+
+    def __init__(self, robot, tf_listener, tf_buffer):
```
Olavhaasie: `tf_listener` is not used, was there a reason for adding this?
```suggestion
    def __init__(self, robot, tf_buffer):
```

march_data_collector/src/march_data_collector/DataCollectorNode.py
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env python
```
Olavhaasie: A shbang is not necessary for module files
```suggestion

```

march_data_collector/src/march_data_collector/CoMCalculator.py
```diff
@@ -0,0 +1,72 @@
...
+        unnecessary_links = []
+        for link in self.links:
+            if self.links[link].inertial is None:
+                unnecessary_links.append(link)
+
+        for link in unnecessary_links:
+            del self.links[link]
+
+        for link in self.links:
+            self.mass += self.links[link].inertial.mass
```
Olavhaasie: You can also write this a lot shorter by using `filter` and `sum` functions from Python:
```python
self.links = dict(filter(lambda (_, l): l.inertial is not None, robot.link_map.items()))
self.mass = sum(l.inertial.mass for (_, l) in self.links.items())
```

Ishadijcks:
Did you implement this yourself or is it taken from somewhere?

RutgerVanBeek:
> Did you implement this yourself or is it taken from somewhere?

It is adapted from https://github.com/bit-bots/center_of_mass/tree/master/src/center_of_mass

march_data_collector/src/march_data_collector/com_calculator.py
```diff
@@ -2,27 +2,18 @@
...
-from urdf_parser_py.urdf import URDF
 from visualization_msgs.msg import Marker
 
 
 class CoMCalculator(object):
 
-    def __init__(self, robot, tf_listener, tf_buffer):
+    def __init__(self, robot, tf_buffer):
         self.mass = 0
         self.links = robot.link_map
```
Olavhaasie: These two lines are no longer necessary :wink:

PR #402, opened 2019-11-20
Add rate of 2Hz to requesting controllers

Olavhaasie:
Closes PM-145

## Description
The state machine would spam the controller manager as fast as it could with requests to list the controllers. This would for unknown reasons cause the hardware interface to not launch correctly. I added a rate of 2Hz, tested it and now it will launch successfully.

## Changes
* Add 2Hz rate

### DIFF
--- a/march_state_machine/src/march_state_machine/states/WaitForRosControlState.py
+++ b/march_state_machine/src/march_state_machine/states/WaitForRosControlState.py
@@ -19,6 +19,7 @@ class WaitForRosControlState(smach.State):
         rospy.logdebug('controller manager found')
 
         req = ListControllersRequest()
+        rate = rospy.Rate(2)
         while True:
             if rospy.get_rostime() > end:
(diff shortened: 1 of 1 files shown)


march_state_machine/src/march_state_machine/states/WaitForRosControlState.py
```diff
@@ -19,6 +19,7 @@ def execute(self, userdata):
         rospy.logdebug('controller manager found')
 
         req = ListControllersRequest()
+        rate = rospy.Rate(2)
```
JorisWeeda: why a rate of 2?

Olavhaasie: I tested a rate of 2 and that worked. We can try faster if it turns out to be too slow

JorisWeeda: allright maybe something to keep in mind for the next test

PR #404, opened 2019-11-20
Feature: migrate import gait files

Roelemans:
Closes PM-148

## Description

Create constructors for the Subgait class that allow importing from file and from ros message.


march_shared_classes/src/march_shared_classes/gait/subgait.py
```diff
@@ -19,6 +25,53 @@ def __init__(self, joints, duration, gait_type='walk_like',
...
+            urdf_joint = cls._get_joint_from_urdf(robot, joint_name)
+            if urdf_joint is None:
+                rospy.logwarn("Not all joints in gait are in robot.")
+                continue
+
+            limits = Limits(urdf_joint.safety_controller.soft_lower_limit,
+                            urdf_joint.safety_controller.soft_upper_limit,
+                            urdf_joint.limit.velocity)
+
+            joint_list.append(cls.joint_class.from_msg(subgait_msg, joint_name, limits, duration))
```
Olavhaasie: You could put this functionality in a separate function. So that you do not have to call `from_msg` from `from_file` to load the setpoints, because now the flow is `file > msg > class`, which is a bit redundant

march_shared_classes/src/march_shared_classes/gait/subgait.py
```diff
@@ -2,9 +2,15 @@
...
 from trajectory_msgs.msg import JointTrajectoryPoint
 from march_shared_resources.msg import Setpoint
+from limits import Limits
+from joint_trajectory import JointTrajectory
+import yaml
+from rospy_message_converter import message_converter
 
 
 class Subgait:
+    joint_class = JointTrajectory
```
Olavhaasie: I assume you did this because of the name collision from the message and the class. You could use the message as `msg.JointTrajectory` to avoid collisions.

march_shared_classes/src/march_shared_classes/gait/joint_trajectory.py
```diff
@@ -1,11 +1,33 @@
+import rospy
+from setpoint import Setpoint
+
+
 class JointTrajectory:
+    setpoint_class = Setpoint
```
Olavhaasie: Why do you do this?

march_shared_classes/src/march_shared_classes/gait/subgait.py
Roelemans: No, I use it so I can override it in child classes, allowing me to use a different class then JointTrajectory there.

march_shared_classes/src/march_shared_classes/gait/joint_trajectory.py
Roelemans: Same as the joint_class

Roelemans:
Main reason is that I just copy pasted the code without changing it. Another reason would be that the limits should never change, so no setters are needed, whereas getters are not really used in any of these classes. That, of course, might be a problem in itself. Not in de scope of this PR to resolve that though.

march_shared_classes/src/march_shared_classes/gait/subgait.py
Roelemans: You need to have access to the class to call cls.joint_class, you could avoid this by adding joint_class as a parameter to the constructor method, but then you would be passing this around many times (since you would also need to pass around setpoint_class in that case).  

In addition, it might still be useful if another part of the code happens to want to transfer a dictionary to subgait directly, which might happen if we use these classes in a location that listens to a topic where trajectories are published.

march_shared_classes/src/march_shared_classes/gait/subgait.py
```diff
@@ -19,40 +18,52 @@ def __init__(self, joints, duration, gait_type='walk_like',
...
+            return None
+        try:
+            gait_name = file_name.split("/")[-3]
+            subgait_name = file_name.split("/")[-2]
+            version = file_name.split("/")[-1].replace(".subgait", "")
+            subgait_dict = yaml.load(open(file_name), Loader=yaml.SafeLoader)
+            if 'gait_type' not in subgait_dict:
+                subgait_dict['gait_type'] = 'walk_like'
+        except Exception as e:
+            rospy.logerr(str(e))
```
JorisWeeda: i would add the type of the error as well as the error itself
```suggestion
            rospy.logerr("Error occured in subgait: {}, {} ".format(type(e), e))
```
This would improve the usage of the error messages

march_shared_classes/src/march_shared_classes/gait/joint_trajectory.py
```diff
@@ -1,17 +1,36 @@
...
 
+    @classmethod
+    def from_dict(cls, subgait_dict, joint_name, limits, duration):
+        joint_trajectory = subgait_dict['trajectory']
+        joint_index = joint_trajectory['joint_names'].index(joint_name)
+
+        setpoints = []
+        for point in joint_trajectory['points']:
+            time = rospy.Duration(point['time_from_start']['secs'], point['time_from_start']['nsecs']).to_sec()
+            setpoints.append(cls.setpoint_class(time, point['positions'][joint_index],
```
JorisWeeda: Why not just use Setpoint? then the creating of setpoint_class can be removed.

Roelemans: Because I wanted to allow child classes to use a different Setpoint class by overriding setpoint_class instead of having to override the entire method.

Roelemans:
Working on that. They will come in a seperate pull request.

PR #405, opened 2019-11-21
Fix/wait for gait server

Olavhaasie:
## Description
This PR removes the `WAIT FOR ROS_CONTROL` state and adds `WAIT FOR GAIT SERVER` state. Since the last PR (#402) did not fully fix the problem. I removed this state since the state machine is not directly dependent on controllers and this makes it easier testable.

## Changes
* Removed `WAIT FOR ROS_CONTROL` 
* Added `WAIT FOR GAIT SERVER` 
* Added `wait_for_service` to unpause service


Olavhaasie:
> It keeps being in the Wait For Ros Control State for me.

That is because the simulation still has to wait for the controllers to load before unpausing

march_state_machine/src/march_state_machine/states/wait_for_gait_server_state.py
```diff
@@ -0,0 +1,22 @@
...
+import rospy
+import smach
+
+from march_shared_resources.msg import GaitNameAction
+
+
+class WaitForGaitServerState(smach.State):
+    """State which waits for the gait server to be available."""
+
+    def __init__(self, timeout=rospy.Duration.from_sec(60)):
```
JorisWeeda: When the constant value of 60 is not defined in a variable I would add a comment to this line which explains why this value is 60. Even better is to follow the PEP8 rules and define them up front: 
https://www.python.org/dev/peps/pep-0008/#constants

march_state_machine/src/march_state_machine/states/wait_for_ros_control_state.py
```diff
@@ -19,7 +19,6 @@ def execute(self, userdata):
         rospy.logdebug('controller manager found')
 
         req = ListControllersRequest()
-        rate = rospy.Rate(2)
```
JorisWeeda: When the constant value of 2 is not defined in a variable I would add a comment to this line which explains why this value is 2. Even better is to follow the PEP8 rules and define them up front: 
https://www.python.org/dev/peps/pep-0008/#constants

march_state_machine/src/march_state_machine/states/wait_for_gait_server_state.py
Olavhaasie: Good point I'll change it

march_state_machine/src/march_state_machine/states/__init__.py
```diff
@@ -0,0 +1,2 @@
+"""Default amount of seconds to wait for action server."""
+DEFAULT_TIMEOUT_SECS = 60
```
JorisWeeda: very nice!

march_state_machine/src/march_state_machine/states/wait_for_ros_control_state.py
```diff
@@ -2,11 +2,13 @@
...
 
+from . import DEFAULT_TIMEOUT_SECS
+
 
 class WaitForRosControlState(smach.State):
     """State which waits for the ros controller manager to load the controllers."""
 
-    def __init__(self, timeout=60):
+    def __init__(self, timeout=rospy.Duration(secs=DEFAULT_TIMEOUT_SECS)):
         self._timeout = rospy.Duration.from_sec(timeout)
```
RutgerVanBeek: I get the error for this line that the right side should be integer type, so the part with the Ros duration seems unnecessary. Like in the wait for gait server.

RutgerVanBeek:
> > It keeps being in the Wait For Ros Control State for me.
> 
> That is because the simulation still has to wait for the controllers to load before unpausing

I still have that on this branch it keeps hanging in the wait for ROS control state. On this branch it happens always on develop it happens sometimes.

PR #407, opened 2019-11-22
Add slope up gait

marnixbrands:
## Description
Add Ramp and Door slope up gait to state machine


march_state_machine/src/march_state_machine/ramp_door_slope_up_sm.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env python
```
Olavhaasie: This is not necessary

march_state_machine/src/march_state_machine/healthy_sm.py
```diff
@@ -109,6 +110,10 @@ def create():
                                transitions={'succeeded': 'STANDING',
                                             'preempted': 'failed', 'failed': 'UNKNOWN'})
 
+        smach.StateMachine.add('GAIT RD SLOPE UP', ramp_door_slope_up_sm.create(), # RD stands for Ramp and Door
```
Olavhaasie: This will fix your build
```suggestion
        smach.StateMachine.add('GAIT RD SLOPE UP', ramp_door_slope_up_sm.create(),  # RD stands for Ramp and Door
```

PR #408, opened 2019-11-22
Feature/con 9 interpolation method

Ciliaclaij:
## Description
This PR adds a new functionality for gain scheduling. The function made can be used to interpolate between the PID values matched to the different gait types.

## Changes

* Interpolate function
* Test functions added


march_gain_scheduling/src/march_gain_scheduling/one_step_linear_interpolation.py
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python
+
+import rospy
+
+
+def interpolate(current_gains, needed_gains, gradient, delta_t):
+    if len(current_gains) != len(needed_gains):
+        rospy.logerr("needed_gains and current_gains not equal length")
+        return current_gains
```
Ishadijcks: Throw an exception here instead of returning incorrect values

march_gain_scheduling/src/march_gain_scheduling/one_step_linear_interpolation.py
```diff
@@ -0,0 +1,21 @@
...
+    next_gains = [0] * len(current_gains)
+    for i in range(len(current_gains)):
+        if current_gains[i] > needed_gains[i]:
+            next_gains[i] = current_gains[i] - gradient * delta_t
+            if next_gains[i] < needed_gains[i]:
+                next_gains[i] = needed_gains[i]
```
Ishadijcks:
```
next_gains = max(needed_gains[i], current_gains[i] - gradient * delta_t)
```

march_gain_scheduling/src/march_gain_scheduling/one_step_linear_interpolation.py
```diff
@@ -0,0 +1,21 @@
...
+        else:
+            next_gains[i] = current_gains[i] + gradient * delta_t
+            if next_gains[i] > needed_gains[i]:
+                next_gains[i] = needed_gains[i]
```
Ishadijcks: Same

march_gain_scheduling/src/march_gain_scheduling/one_step_linear_interpolation.py
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python
```
Olavhaasie: this can be removed

march_gain_scheduling/src/march_gain_scheduling/one_step_linear_interpolation.py
Olavhaasie: It is better to not log anything from library methods, then the user can determine whether to log or not. Furthermore, You should return `None` or raise an exception, otherwise it is hard to tell whether the returned array is the interpolated version or the first argument.

march_gain_scheduling/test/one_step_linear_interpolation_test.py
```diff
@@ -0,0 +1,44 @@
...
+        needed = [4, 3, 4]
+        while input != output:
+            result = interpolate(current, needed, 1, 0.1)
+        self.assertEqual(result, [4, 3, 4])
+
+    def test_interpolate_list_length(self):
+        current = [2, 5, 6, 8]
+        needed = [4, 3, 4]
+        with self.assertRaises(TypeError):
+            interpolate(current, needed, 1, 0.1)
```
Olavhaasie: This should not succeed :thinking:

march_gain_scheduling/src/march_gain_scheduling/one_step_linear_interpolation.py
Deakensbubbles: next_gains[i] = max(needed_gains[i], current_gains[i] - gradient * delta_t)

Subtle difference, but works better. also, using this method as suggested by Isha is neater, but requires an extra check for gains that are equal to the previous gains.

march_gain_scheduling/src/march_gain_scheduling/one_step_linear_interpolation.py
Deakensbubbles: another subtle difference: use min() instead of max() ergo: next_gains[i] = min(needed_gains[i], current_gains[i] + gradient * delta_t)

march_gain_scheduling/src/march_gain_scheduling/one_step_linear_interpolation.py
Roelemans: It is better to just throw an exception here. When this happens there is a significant problem with how this function is used. In the code this function is used, one can catch this exception (i.e. with try and except https://www.pythonforbeginners.com/error-handling/python-try-and-except/) and prevent the exo from shutting down. Its probably best to make your own exception (@Olavhaasie?) see https://www.programiz.com/python-programming/user-defined-exception

march_gain_scheduling/test/one_step_linear_interpolation_test.py
```diff
@@ -0,0 +1,44 @@
...
+        current = [5, 5, 5]
+        needed = [3, 3, 3]
+        result = interpolate(current, needed, 1, 0.1)
+        self.assertEqual(result, [4.9, 4.9, 4.9])
+
+    def test_interpolate_up_and_down(self):
+        current = [2, 5, 6]
+        needed = [4, 3, 4]
+        while input != output:
+            result = interpolate(current, needed, 1, 0.1)
```
Roelemans: If this does not work because someone broke your code, this loop will be infinite. Maybe nice to change it so that it iterates a maximum number of times :innocent:

march_gain_scheduling/test/one_step_linear_interpolation_test.py
Roelemans: Nice error testing :+1: unfortunate that this test will fail :P, but if you change the code so that it throws an error as mentioned above, you will only need to change the type of error checked for here :)

march_gain_scheduling/test/one_step_linear_interpolation_test.py
```diff
@@ -0,0 +1,44 @@
...
+        needed = [4, 3, 4]
+        with self.assertRaises(TypeError):
+            interpolate(current, needed, 1, 0.1)
+
+
+    def test_interpolate_different_paths(self):
+        current = [1, 1, 1]
+        needed = [2, 5, 4]
+        while current != needed:
+            result = interpolate(current, needed, 1, 0.1)
```
Roelemans: Same as above. Try to avoid infinite loops :)

march_gain_scheduling/test/one_step_linear_interpolation_test.py
```diff
@@ -0,0 +1,44 @@
...
+        self.assertEqual(result, [2, 5, 4])
+
+
```
Roelemans: Lastly, maybe nice to add one more test where you use a gradient and a delta_t that are not 1 and 0.1? If you accidentally hard-code these values in your function, shit is gonna go down :)

Ishadijcks:
> Nice work! Your code is probably better tested than any other part of the code.

What if both lists are empty, or if I don't pass lists but strings ;)

march_gain_scheduling/test/one_step_linear_interpolation_test.py
JorisWeeda: and also "output" is not defined right?

march_gain_scheduling/test/one_step_linear_interpolation_test.py
JorisWeeda: I agree, or put a check in the function to verify the input values.

march_gain_scheduling/src/march_gain_scheduling/one_step_linear_interpolation.py
Olavhaasie: Probably best to define your own

Ciliaclaij: I am trying to implement this into the code, but I can't figure out how.
This was my first guess:
`try:
        for i in range(len(current_gains)):
            if current_gains[i] > needed_gains[i]:
                 next_gains[i] = max(needed_gains[i], current_gains[i] - gradient * delta_t)
            else:
                 next_gains[i] = min(needed_gains[i], current_gains[i] - gradient * delta_t)
except:
           ....  
`
The problem when using this implementation is that you cover the situation where current_gains is larger than needed_gains, but not the other way around. Because that implementation will not make the loop fail.. Or do I need to make a implementation like this:

march_gain_scheduling/src/march_gain_scheduling/unequal_length_error.py
```diff
@@ -0,0 +1,5 @@
+import rospy
+
+class UnequalLengthError(Exception):
```
JorisWeeda: When creating a customized error it is nice to add some additional information:
```suggestion
class UnequalLengthError(Exception):
    def __init__(self, msg=None):
        """Error class for the unequal length situation

        :param msg:
            The message that needs to be displayed.
        """
        if msg is None:
            msg = 'current_gains and needed_gains do not have the same length' 
         super(UnequalLengthError, self).__init__(msg)
```

march_gain_scheduling/src/march_gain_scheduling/unequal_length_error.py
```diff
@@ -0,0 +1,5 @@
...
+    print "current_gains and needed_gains do not have the same length"
```
JorisWeeda: And instead of printing maybe logging this error is better.

march_gain_scheduling/src/march_gain_scheduling/unequal_length_error.py
Ishadijcks: `UnequalLengthError` implies it is generic across all arrays, yet its message is specific for current and needed gains

march_gain_scheduling/test/one_step_linear_interpolation_test.py
```diff
@@ -21,23 +21,21 @@ def test_interpolate_down(self):
...
-
+        self.assertRaises(Exception, interpolate, current, needed, 1, 0.1)
 
     def test_interpolate_different_paths(self):
         current = [1, 1, 1]
         needed = [2, 5, 4]
-        while current != needed:
-            result = interpolate(current, needed, 1, 0.1)
-        self.assertEqual(result, [2, 5, 4])
+        for i in range(30):
```
Ishadijcks: 30?

march_gain_scheduling/test/one_step_linear_interpolation_test.py
Olavhaasie: Perhaps also add a test for empty lists?

march_gain_scheduling/src/march_gain_scheduling/unequal_length_error.py
Ciliaclaij: Removed print

march_gain_scheduling/src/march_gain_scheduling/unequal_length_error.py
Ciliaclaij: I am going to leave the exception empty. A message is passed with more explanation when it is called.

march_gain_scheduling/test/one_step_linear_interpolation_test.py
Ciliaclaij: Added both tests :+1:

march_gain_scheduling/test/one_step_linear_interpolation_test.py
Ciliaclaij: Added explanation

march_gain_scheduling/test/one_step_linear_interpolation_test.py
```diff
@@ -0,0 +1,44 @@
...
+        needed = [3, 3, 3]
+        result = interpolate(current, needed, 1, 0.1)
+        self.assertEqual(result, [4.9, 4.9, 4.9])
+
+    def test_interpolate_up_and_down(self):
+        current = [2, 5, 6]
+        needed = [4, 3, 4]
+        for i in range(25):
+            current = interpolate(current, needed, 3, 0.1)
+        self.assertEqual(current, [4, 3, 4])
```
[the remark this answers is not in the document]
Ciliaclaij: Added Explanation

march_gain_scheduling/test/one_step_linear_interpolation_test.py
```diff
@@ -0,0 +1,57 @@
...
+            interpolate(current, needed, -1, 0.1)
+
+    def test_interpolate_different_paths(self):
+        current = [1, 1, 1]
+        needed = [2, 5, 4]
+        for i in range(26):  # Loop until needed is reached and a bit further
+            current = interpolate(current, needed, 1, 0.2)
+        self.assertEqual(current, [2, 5, 4])
+
+    def test_interpolate_up(self):
```
Olavhaasie: This test name is the same as the first test (probably because you copied it :speak_no_evil:)
```suggestion
    def test_interpolate_empty_lists(self):
```

march_gain_scheduling/test/one_step_linear_interpolation_test.py
```diff
@@ -0,0 +1,57 @@
...
+        current = []
+        needed = []
+        result = interpolate(current, needed, 1, 0.1)
+        self.assertEqual(result, [])
+
+
+
+if __name__ == '__main__':
+    rosunit.unitrun(PKG, 'test_multiply', OneStepLinearInterpolationTest)
```
Olavhaasie: `PKG` is undefined, you can solve this by adding 
```python
PKG = 'march_gain_scheduling`
```
At the top of the file, like they do in the [examples](https://wiki.ros.org/unittest#Code-level_Python_Unit_Tests).

march_gain_scheduling/test/one_step_linear_interpolation_test.py
Ciliaclaij: Fixed

Roelemans: PS this was a mistake in our tutorials, it has now been fixed

march_gain_scheduling/test/one_step_linear_interpolation_test.py
```diff
@@ -0,0 +1,58 @@
...
+    def test_negative_value_error(self):
+        current = [2, 5, 6]
+        needed = [4, 3, 4]
+        with self.assertRaises(NegativeValueError):
+            interpolate(current, needed, -1, 0.1)
+
+    def test_interpolate_different_paths(self):
+        current = [1, 1, 1]
+        needed = [2, 5, 4]
+        for i in range(26):  # Loop until needed is reached and a bit further
```
Ishadijcks: For this test to have any meaning you would need to confirm that it is not equal in earlier loops

PR #412, opened 2019-11-26
Feature: implement state machine shutdown

Olavhaasie:
Closes PM-61

## Description
This PR fixes the preemption of the statemachine correctly when shutting down ROS. When ROS is shutdown, the state machine will preempt the current states and shutdown gracefully by going to the `SHUTDOWN` state. The concurrent `SafetyState` has now been implemented as a `smach_ros.MonitorState`, which will preempt the state machine when something is published on `/march/error`. Then the state machine will go to `ERROR`, which is currently not implemented and automatically goes to `SHUTDOWN`. In the future this can be used for any error handling to return to `HEALTHY`. The state machine is now also capable of preempting the `WAIT FOR GAIT SERVER` state. Furthermore, the `IdleState` is now implemented with callbacks from `control_flow`, which made it easier to preempt it.

**tl;dr** No more errors when pressing ctrl-c :ok_man: 

## Changes
* Implement `SafetyState` as `MonitorState`
* Add preemption of concurrent states
* Add preemption on `IdleState`
* Add `ShutdownState`
* Some cleaning up of code that was no longer necessary
* Renaming files from CamelCase to underscore_case


march_state_machine/src/march_state_machine/states/idle_state.py
```diff
@@ -0,0 +1,59 @@
...
+        control_flow.reset_stop()
+
+    def _gait_cb(self, gait):
+        if gait in self.get_registered_outcomes():
+            rospy.loginfo('Accepted {}'.format(gait))
+            self._result_gait = gait
+            control_flow.gait_accepted()
+            self._trigger_event.set()
+        else:
+            rospy.logwarn('The %s is not a possible gait in the current state', result_gait)
```
RutgerVanBeek: The last argument should be gait, because result_gait is not necessarily set.

march_state_machine/src/march_state_machine/states/shutdown_state.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class ShutdownState(smach.State):
+    """Shutdown state that will request shutdown from ROS. Will always succeed."""
+
+    def __init__(self):
+        smach.State.__init__(self, outcomes=['succeeded'])
+
+    def execute(self, _userdata):
+        rospy.signal_shutdown('State machine shutdown')
```
RutgerVanBeek: Aren't there actions to take for gracefully shutting down, like shutting down gazebo for instance?

Olavhaasie: I was not able to find a way to gracefully shutdown gazebo. When `SIGINT` is sent to gazebo it does not seem to shutdown, the only way to shut it down is by sending `SIGTERM`

RutgerVanBeek: Do we break something in Gazebo by sending that directly?

Olavhaasie: No probably not, but its not a clean solution, there is an issue open for the issue https://github.com/ros-simulation/gazebo_ros_pkgs/issues/751

march_state_machine/src/march_state_machine/control_flow.py
```diff
@@ -14,6 +14,27 @@ def __init__(self):
...
+
+    def clear_callbacks(self):
+        """Clears all currently registered callbacks."""
+        self._stopped_cb = None
+        self._gait_selected_cb = None
+
+    def set_stopped_callback(self, cb):
+        """Sets the callback for when stop is pressed.
+
+        :param cb: Callback that will be called with no arguments
```
JorisWeeda: nice, docstrings!!

march_state_machine/src/march_state_machine/states/shutdown_state.py
Roelemans: Please implement an ugly solution then

Olavhaasie:
Don't forget to install `psutil`

    rosdep install -y --from-paths src --ignore-src --rosdistro kinetic

march_state_machine/src/march_state_machine/control_flow.py
```diff
@@ -14,6 +14,27 @@ def __init__(self):
...
+        """
+        self._stopped_cb = cb
+
+    def set_gait_selected_callback(self, cb):
+        """Sets the callback for when a new gait is selected.
+
+        :param cb: Callback that will be called with one string argument gait_name
+        """
+        self._gait_selected_cb = cb
```
Roelemans: Why do you use setters here? Why not make these variables non-private and directly edit the from idle_state.py?

march_state_machine/src/march_state_machine/states/safety_state.py
```diff
@@ -0,0 +1,29 @@
...
+class SafetyState(smach_ros.MonitorState):
+    """State which monitors if errors have occurred.
+
+    This state always runs concurrent to the healthy state machine.
+    Listens to the topic '/march/error' and returns 'invalid' when
+    an error occurs and 'valid' when no error has occurred.
+    """
+
+    def __init__(self):
+        smach_ros.MonitorState.__init__(self, '/march/error', Error, SafetyState.error_callback)
```
Roelemans:
```suggestion
        super(SafetyState, self).__init__('/march/error', Error, self.error_callback)
```

march_state_machine/src/march_state_machine/healthy_sm.py
```diff
@@ -47,79 +47,75 @@ def create():
...
+                               transitions={'succeeded': 'STANDING', 'failed': 'UNKNOWN'})
 
         smach.StateMachine.add('GAIT RT MIDDLE STEPS', rough_terrain_middle_steps_sm.create(),
-                               transitions={'succeeded': 'STANDING',
-                                            'preempted': 'failed', 'failed': 'UNKNOWN'})
+                               transitions={'succeeded': 'STANDING', 'failed': 'UNKNOWN'})
 
         smach.StateMachine.add('GAIT RD SLOPE UP', ramp_door_slope_up_sm.create(),  # RD stands for Ramp and Door
                                transitions={'succeeded': 'STANDING',
                                             'preempted': 'failed', 'failed': 'UNKNOWN'})
```
Roelemans: Remove preempted here

march_state_machine/src/march_state_machine/states/idle_state.py
```diff
@@ -0,0 +1,59 @@
...
+        if self._result_gait:
+            return self._result_gait
+
+        return 'failed'
+
+    def _stopped_cb(self):
+        rospy.logwarn('Idle state doesn\'t respond to stop')
+        control_flow.reset_stop()
+
+    def _gait_cb(self, gait):
```
Roelemans: Are these functions really private if they are connected to the callback of another class?

march_state_machine/src/march_state_machine/states/idle_state.py
```diff
@@ -0,0 +1,59 @@
...
+import rospy
+import threading
+from march_state_machine.control_flow import control_flow
+
+
+class IdleState(smach.State):
+    """State in which the exoskeleton is not moving.
+    Listens to instructions from the input device and reacts if they are known transitions.
+    """
+    def __init__(self, outcomes=None):
```
Roelemans: Why not default outcomes to []?

march_state_machine/src/march_state_machine/states/shutdown_state.py
```diff
@@ -0,0 +1,23 @@
...
+
+        # Please forgive me for my sins
+        # This is an issue with gazebo that will not respond to SIGINT
+        # So we send a SIGTERM instead to kill it
+        # See https://github.com/ros-simulation/gazebo_ros_pkgs/issues/751
+        for p in psutil.process_iter():
+            if p.name() == 'gzserver':
+                p.kill()
```
Roelemans: YAAAASSS HELD

march_state_machine/test/rostest/test_state_machine.py
```diff
@@ -4,7 +4,7 @@
 import rospy
 from smach import InvalidTransitionError, InvalidStateError
 
-import march_state_machine.StateMachine as StateMachine
+import march_state_machine.state_machine as sm
```
Roelemans: Why this? You could also just do `from march_state_machine import StateMachine`

Olavhaasie: Except `StateMachine` is not a class, so that won't work

march_state_machine/src/march_state_machine/states/idle_state.py
Olavhaasie: The default argument value of `[]` is mutable in Python, which can do weird things apparently. See https://stackoverflow.com/questions/366422/what-is-the-pythonic-way-to-avoid-default-parameters-that-are-empty-lists

march_state_machine/src/march_state_machine/states/idle_state.py
Olavhaasie: They are not necessarily private more like 'not public'. These callbacks are not supposed to be called from outside the class, which will give unexpected behaviour. That's why I chose to make them not public.

march_state_machine/src/march_state_machine/control_flow.py
Olavhaasie: I found this more readable and has a clearer intention

march_state_machine/src/march_state_machine/states/idle_state.py
Roelemans: Cool

march_state_machine/test/rostest/test_state_machine.py
Roelemans: Ah I meant `march_state_machine as state_machine`, but then I guess you might as well do this then.

march_state_machine/src/march_state_machine/states/stoppable_state.py
```diff
@@ -1,8 +1,8 @@
...
 
-from FeedbackActionState import FeedbackActionState
-from march_shared_resources.msg import GaitNameAction, GaitNameGoal, GaitInstruction
-from march_state_machine.ControlFlow import control_flow
+from feedback_action_state import FeedbackActionState
+from march_shared_resources.msg import GaitNameAction, GaitNameGoal
+from march_state_machine.control_flow import control_flow
 
 
 class StoppableState(FeedbackActionState):
```
Roelemans: I think this class should have a similar implementation of control_flow to the one in IdleState. This will also allow the removal of some unused methods in control_flow.

Olavhaasie: This class is a bit different in implementation since it inherits from `SimpleActionState`, which implements preemption for itself, so it is not possible to use the same kind of implementation.

PR #416, opened 2019-12-02
Feature/blc 16 implementing cp

Deakensbubbles:
Closes BLC-16

## Description

Add capture point markers to the march_data_collector for visualization in RVIZ. Don't forget to add the markers in the ui.
Calculating Capture Point is the first step in providing a balanced gait on the exoskeleton.

## Changes

Added CPCalculator class and usage of it in data_collector_node
Changed trajectory_state_callback to update the markers.


march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -0,0 +1,54 @@
...
+        self.marker.action = self.marker.ADD
+        self.marker.pose.orientation.w = 1.0
+        self.marker.color.a = 1.0
+        self.marker.color.g = 1.0
+        self.marker.scale.x = 0.03
+        self.marker.scale.y = 0.03
+        self.marker.scale.z = 0.03
+
+        g = 9.81  # gravity constant
+        z_zero = 0.715500019353  # height of CoM
```
Ishadijcks: Hardcoded?

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -0,0 +1,54 @@
...
+        time_difference = (current_time - self.prev_t).to_sec()
+        if current_time is not self.prev_t:
+            x_dot = (com_mark.pose.position.x - self.prev_x) / time_difference
+            y_dot = (com_mark.pose.position.y - self.prev_y) / time_difference
+
+            trans = self.tf_buffer.lookup_transform('world', self.foot_link, rospy.Time())
+            x_cap = trans.transform.translation.x + x_dot * self.MULTIPLICATION_CONSTANT
+            y_cap = trans.transform.translation.y + y_dot * self.MULTIPLICATION_CONSTANT
+
+            # send CP position to RViZ
```
Ishadijcks: Extract to different method

march_data_collector/src/march_data_collector/cp_calculator.py
Deakensbubbles: Jep, this value (z_zero) isn't going to change (much). The movement of the CoM is more important for the calculations.

Ishadijcks: How is it calculated?

Deakensbubbles: It is the value of the CoM when the exo is standing minus the hover height (if there is one). This could be implemented later on when using the exo, but for now it's fine to use this constant.

Ishadijcks: > It is the value of the CoM when the exo is standing

How is this calculated?

march_data_collector/src/march_data_collector/data_collector_node.py
```diff
@@ -12,16 +12,21 @@
...
-    def __init__(self, com_calculator):
+    def __init__(self, com_calculator, cp_calculators):
         self._com_calculator = com_calculator
+        self._cp_calculators = cp_calculators
         joint_names = rospy.get_param('/march/joint_names')
         self._imu_broadcaster = tf2_ros.TransformBroadcaster()
 
         self._marker_publisher = rospy.Publisher('/march/com_marker', Marker, queue_size=1)
+        self._com_marker_publisher = rospy.Publisher('/march/com_marker', Marker, queue_size=1)
+        self._cp_marker_publishers = [rospy.Publisher('/march/cp_marker_left', Marker, queue_size=1),
```
Olavhaasie: You expect a list of cp calculators, but you only define two publishers here. So either only accept two `cp_right` and `cp_left` or create the publishers from the given `cp_calculators`

march_data_collector/src/march_data_collector/data_collector_node.py
```diff
@@ -63,16 +71,18 @@ def imu_callback(self, data):
             transform.transform.rotation.z = imu_rotation[2]
             transform.transform.rotation.w = imu_rotation[3]
 
-            self._imu_broadcaster.sendTransform(transform)
+            self.imu_broadcaster.sendTransform(transform)
```
Olavhaasie: `self.imu_broadcaster` is not a defined value
```suggestion
            self._imu_broadcaster.sendTransform(transform)
```

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -0,0 +1,54 @@
...
+        self.MULTIPLICATION_CONSTANT = sqrt(z_zero / g)
+
+    def calculate_cp(self, com_mark):
+        current_time = com_mark.header.stamp
+        time_difference = (current_time - self.prev_t).to_sec()
+        if current_time is not self.prev_t:
+            x_dot = (com_mark.pose.position.x - self.prev_x) / time_difference
+            y_dot = (com_mark.pose.position.y - self.prev_y) / time_difference
+
+            trans = self.tf_buffer.lookup_transform('world', self.foot_link, rospy.Time())
```
Olavhaasie: The default `rospy.Time()`[ constructor initializes on 0 seconds](https://docs.ros.org/kinetic/api/rospy/html/rospy.rostime.Time-class.html#__init__), which might not be what you want.
```suggestion
            trans = self.tf_buffer.lookup_transform('world', self.foot_link, rospy.get_rostime())
```

march_data_collector/src/march_data_collector/cp_calculator.py
Olavhaasie: The comment is also wrong since you do not send anything. I would create a method `update_marker` that updates the position and time of the marker.

march_data_collector/src/march_data_collector/cp_calculator.py
Roelemans: I think Isha is trying to say that is not very neat to hard-code this value, but instead it should be calculated from the URDF description. Perhaps you could pass robot tot this method and calculate the value from there? That way, shit won't break when I start messing around with the lengths and masses of the exo again :innocent: 

Edit: Perhaps you can very simply use the com_mark below to dynaimcally update it?

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -0,0 +1,54 @@
...
+            x_cap = trans.transform.translation.x + x_dot * self.MULTIPLICATION_CONSTANT
+            y_cap = trans.transform.translation.y + y_dot * self.MULTIPLICATION_CONSTANT
```
Roelemans: Call these x_cp and y_cp for better consistency :)

Roelemans: When using capture point, don't forget to take into account that this only holds when z_dot is small. Hence, you should not look at the capture point at the start of a step, and not at its changes during a step.

march_data_collector/src/march_data_collector/cp_calculator.py
RutgerVanBeek: I am going to use the com_marker. This makes sure that changes in the URDF are taken into account. However it is not ideal as for instance when you are sitting the z_0 is much lower and the capture point doesn't make much sense anymore.

march_data_collector/src/march_data_collector/cp_calculator.py
RutgerVanBeek: Agree

march_data_collector/src/march_data_collector/cp_calculator.py
RutgerVanBeek: By setting the time argument to zero, it will get the latest and that is exactly what we want.
http://docs.ros.org/indigo/api/tf2_ros/html/c++/classtf2__ros_1_1Buffer.html#acabbd72cae8f49fb3b6ede3be7e34c55

RutgerVanBeek:
> What does this implementation currently show in RViz? I currently see green dots that are always under the shoes

The capture point is the point where you should step in order to be able to stay completely stable. However as long as the exoskeleton is fixed these points do indeed not make a lot of sense. This the main reason for the other PR to be able to unfix the exo in Rviz.

RutgerVanBeek:
> Also when I launch the node I get these warnings. Any idea what they mean?
> 
> ```
> [WARN] error in CoM calculation"world" passed to lookupTransform argument target_frame does not exist. 
> [WARN] error in CoM calculation"world" passed to lookupTransform argument target_frame does not exist. 
> [WARN] error in CoM calculation"world" passed to lookupTransform argument target_frame does not exist. 
> [WARN] error in CoM calculation"world" passed to lookupTransform argument target_frame does not exist. 
> [WARN] error in CoM calculationCould not find a connection between 'world' and 'foot_left' because they are not part of the same tree.Tf has two or more unconnected trees.
> [WARN] error in CoM calculationCould not find a connection between 'world' and 'foot_right' because they are not part of the same tree.Tf has two or more unconnected trees.
> [WARN] error in CoM calculation"lower_leg_left" passed to lookupTransform argument source_frame does not exist. 
> [WARN] error in CoM calculationCould not find a connection between 'world' and 'hip_aa_frame_left_front' because they are not part of the same tree.Tf has two or more unconnected trees.
> ```

The cpCalculator was initiated with a center of mass. This was requested before the ComCalculator was fully ready. I have changed the initiation of the CPcalculator to not ask for the com for initiation.

PR #418, opened 2019-12-03
Feature/con 61 implementation interpolation

Ciliaclaij:
Closes CON-61


Ishadijcks:
Please squash this pr when merging

march_gain_scheduling/src/march_gain_scheduling/one_step_linear_interpolation.py
```diff
@@ -12,5 +13,6 @@ def interpolate(current_gains, needed_gains, gradient, delta_t):
             next_gains[i] = max(needed_gains[i], current_gains[i] - gradient * delta_t)
         else:
             next_gains[i] = min(needed_gains[i], current_gains[i] + gradient * delta_t)
-
-    return next_gains
+    if current_gains == needed_gains:
```
Ishadijcks: Shouldn't this be `next_gains`?

march_gain_scheduling/src/march_gain_scheduling/one_step_linear_interpolation.py
```diff
@@ -1,7 +1,8 @@
-from .interpolation_errors import NegativeValueError, UnequalLengthError
+from .interpolation_errors import UnequalLengthError, NegativeValueError
 
 
 def interpolate(current_gains, needed_gains, gradient, delta_t):
+    interpolation_done = False
```
Ishadijcks: Why do you need this value?

Ciliaclaij: We need it to confirm that the interpolation is still needed. So it will only use the interpolation when either the gait_type changes or the interpolation is not done yet (because the gait type just changed). This can be seen in the dynamic_pid_reconfigurer file line 25.

march_gain_scheduling/launch/march_gain_scheduling.launch
```diff
@@ -1,6 +1,12 @@
 <!--change march4 to the necessary use of the package i.e. when using the test_joint use the corresponding folder instead.-->
 <launch>
     <arg name="march_type" default="march4"/>
+    <arg name="linear" default="true"/>
+    <arg name="slope" default="10.0"/>
+
+    <param name="linearize_gain_scheduling" type="bool" value="$(arg linear)"/>
+    <param name="linear_slope" type="double" value="$(arg slope)"/>
```
Olavhaasie: It would be nicer to add these parameters as private parameters to the `gain_scheduling_node` like so:
```
<node name="gain_scheduling_node" pkg="march_gain_scheduling" type="march_gain_scheduling_node" output="screen">
    <rosparam command="load" file="$(find march_gain_scheduling)/config/$(arg march_type)/gains_per_gait_type.yaml"/>
    <param name="linearize_gain_scheduling" type="bool" value="$(arg linear)"/>
    <param name="linear_slope" type="double" value="$(arg slope)"/>
</node>
```

Then you can request the parameters using  `rospy.get_param('~linearize_gain_scheduling')` and `rospy.get_param('~linear_slope')`

march_gain_scheduling/src/march_gain_scheduling/dynamic_pid_reconfigurer.py
```diff
@@ -1,39 +1,63 @@
...
     def client_update(self):
-        for i in range(len(self._joint_list)):
-            p_value, i_value, d_value = self.look_up_table(i)
-            if p_value is not None:
-                self._clients[i].update_configuration({'p': p_value,
-                                                       'i': i_value,
-                                                       'd': d_value})
-                rospy.logdebug('Config set to {0}, {1}, {2}'.format(p_value, i_value, d_value))
+        rospy.logdebug('self.linearize is: {0}'.format(self._linearize))
+        if self._linearize:
```
Olavhaasie: The value of `self._linearize` is nowhere changed, so it will always continue to linearize and never schedule new gains.

march_gain_scheduling/src/march_gain_scheduling/dynamic_pid_reconfigurer.py
```diff
@@ -1,39 +1,63 @@
...
-        if rospy.has_param('~gait_types/' + self._gait_name + '_gains/'):
-            gains = rospy.get_param('~gait_types/' + self._gait_name + '_gains/' + self._joint_list[i])
-            return gains['p'], gains['i'], gains['d']
+    def look_up_table(self, joint_index):
+        if rospy.has_param('~gait_types/' + self._gait_type):
+            gains = rospy.get_param('~gait_types/' + self._gait_type + '/' + self._joint_list[joint_index])
+            return [gains['p'], gains['i'], gains['d']]
         else:
-            return None, None, None
+            return [None, None, None]
```
Olavhaasie: Why did you change this to an array?

march_gain_scheduling/src/march_gain_scheduling/one_step_linear_interpolation.py
```diff
@@ -12,5 +13,6 @@ def interpolate(current_gains, needed_gains, gradient, delta_t):
             next_gains[i] = max(needed_gains[i], current_gains[i] - gradient * delta_t)
         else:
             next_gains[i] = min(needed_gains[i], current_gains[i] + gradient * delta_t)
-
-    return next_gains
+    if current_gains == needed_gains:
+        interpolation_done = True
+    return next_gains, interpolation_done
```
Olavhaasie: You could remove the `interpolation_done` variable and instead return this
```python
return next_gains, next_gains == needed_gains
```

march_gain_scheduling/src/march_gain_scheduling/dynamic_pid_reconfigurer.py
Ciliaclaij: It is supposed to be a choice before initializing. So you can choose beforehand whether you want to use linearization or not.

march_gain_scheduling/src/march_gain_scheduling/dynamic_pid_reconfigurer.py
Ciliaclaij: @Roelemans ?

Roelemans: Because the output is used in interpolate(...) which expects a list. Futhermore, I think it makes more sense to return a list than three separate values.

march_gain_scheduling/src/march_gain_scheduling/one_step_linear_interpolation.py
Roelemans: I think it more neat to move this functionality outside of this method. i.e. check in dynamic_pid_reconfigurer.py after line 39 whether the output of interpolate is equal to needed_gains. That way, the interpolation method has a single, clear 
functionality :D

march_gain_scheduling/src/march_gain_scheduling/dynamic_pid_reconfigurer.py
```diff
@@ -26,7 +26,10 @@ def gait_selection_callback(self, data):
             rospy.logdebug('The selected gait: {0} is not the same as the previous gait: {1}'.format(
                 data.goal.current_subgait.gait_type, self._gait_type))
             self._gait_type = data.goal.current_subgait.gait_type
-            self.client_update()
+            self.interpolation_done = False
+            while not self.interpolation_done:
+                self.interpolation_done = True
```
Roelemans: I think this is set to true by client.update when it it finished? So this line shouldn't be necessary.

march_gain_scheduling/src/march_gain_scheduling/dynamic_pid_reconfigurer.py
```diff
@@ -26,7 +26,10 @@ def gait_selection_callback(self, data):
...
+                self.client_update()
```
Roelemans: Use `rate = rospy.Rate(Hz)` above the loop and `rate.sleep()` inside the loop so that it is called at rate Hz.

Edit: use try and except around the rate.sleep() as Olav suggested in https://github.com/project-march/simulation/pull/34

Ciliaclaij: This needs to be done because you only need to update after a few milliseconds?

Olavhaasie: This loop will run as fast as possible and will therefore interpolate with very small time steps

Deakensbubbles: will the removal of line 31 not cause the callback to loop infinitely if self._linearize is set to False?

Roelemans: Ah, I understand better now. In that case it is probably more need to set self.interpolation_done to true within the else statement of client_update. This way, client_update decides when the interpolation is done.

Roelemans:
Hey @Ciliaclaij, I took the freedom to make a small fix since you seem are very busy with important IMC stuff :). Just realised you can compare current_gains to needed_gains, removing the need for the extra variable next_gains. Also, in the previous implementation, interpolation_done was set to true when a single joint was finished, while this should wait untill all are done. Fixed that as well :).

There is one more small style improvement possible. If you want that, you can come by when you have time so I can show you.

march_gain_scheduling/src/march_gain_scheduling/dynamic_pid_reconfigurer.py
```diff
@@ -38,16 +40,19 @@ def client_update(self):
...
-                self._clients[i].update_configuration({'p': self.current_gains[i][0],
-                                                       'i': self.current_gains[i][1],
-                                                       'd': self.current_gains[i][2]})
-                rospy.logdebug('Config set to {0}, {1}, {2}'.format(self.current_gains[i][0],
-                                                                    self.current_gains[i][1],
-                                                                    self.current_gains[i][2]))
+                self.next_gains = [0] * len(needed_gains)
+                self.next_gains[i] = interpolate(self.current_gains[i], needed_gains, gradient, time_interval)
+                if self.next_gains[i] == self.current_gains:
+                    self.interpolation_done = True
```
Olavhaasie: You initialize the next gains here as an array with the same amount of elements as `needed_gains` which is always equal to 3. Then you address this array using the `i` variable which represents the current joint index which could be more than 3. I think the `self.current_gains` variable is correctly initialized. Finally the if statement compares an element with with a 2d array of elements. Therefore the interpolation will never finish.

I think the `next_gains` variable should not be bound to `self` since it used locally in this method. and you only want to stop with the interpolation until every joint has finished interpolating right? So the if statement should be outside of the loop and check whether all `next_gains` are equal to the all the `needed_gains`.

Ciliaclaij: Roel already fixed this

march_launch/launch/march.launch
```diff
@@ -43,6 +43,15 @@
...
 
+    <group if="$(eval configuration == 'test_joint_rotational')">
+        <include file="$(find march_hardware_interface)/launch/hardware.launch" if="$(arg hardware_interface)">
+            <arg name="robot" value="test_joint_rotational"/>
+        </include>
+        <include file="$(find march_gain_scheduling)/launch/march_gain_scheduling.launch">
+             <arg name="march_type" value="test_joint_rotational"/>
+        </include>
+    </group>
+
```
Olavhaasie: This does not really belong here, since it launches a test joint from the high level launch, which also launches all other nodes like the state machine and gait servers. These however have no use when the test joint is used. So you should remove this and if you want to keep it you can put it inside the `march_gain_scheduling.launch` as an option.

Ciliaclaij: I think @Deakensbubbles did this, I actually don't know if we want to keep this. So Rogier could you take a look at this?

Deakensbubbles: I will move this to the `march_gain_scheduling.launch` file as an option.

march_gain_scheduling/launch/march_gain_scheduling.launch
```diff
@@ -1,7 +1,35 @@
...
+
+    <group if="$(eval configuration == 'test_joint_rotational')">
+        <include file="$(find march_launch)/launch/march.launch">
+            <!--This value is not used in the launch file and therefore allows us to start the test joint ourselves.-->
+            <arg name="configuration" value="$(arg configuration)"/>
+        </include>
+        <include file="$(find march_hardware_interface)/launch/hardware.launch">
+            <arg name="robot" value="$(arg configuration)"/>
+        </include>
+    </group>
```
Olavhaasie: > This does not really belong here, since it launches a test joint from the high level launch, which also launches all other nodes like the state machine and gait servers. These however have no use when the test joint is used. So you should remove this and if you want to keep it you can put it inside the march_gain_scheduling.launch as an option.

What I actually meant was to add an option to launch the test joint and only the test joint. The options for `exoskeleton` and `simulation` can be removed, since the `march.launch` should be launching this launch file not the other way around. You should remove the `include` to `march.launch` and only launch the hardware interface. This is because the march network does not work when launching a test configuration.

Roelemans: So more concrete: 
1) Remove line 7 t/m 18 above
2) Remove line 20 t/m 23 above

- The test joint can then be started with gain scheduling by launching this file
- The exoskeleton can be launched with gain scheduling (either in simulation or the real one) by first launching march.launch with the appropriate settings and then launching this file in another terminal. An option to start the exo with gain scheduling can later be added to march.launch (which would then launch this file).

Deakensbubbles: Ah right, so in the `march_gain_scheduling.launch` file should be an include to the hardware_interface.launch and that's it. Then later when gain scheduling is deemed safe enough it will be added as an option to the main march.launch file?

Deakensbubbles: Also, I think I was able to launch the `march.launch` file with an adjusted parameter for the test joint. This was necessary for testing gain scheduling by changing the gait.

Deakensbubbles: @Roelemans The gait selector and rqt input device are both necessary for this node. If I want to adhere to the structure you and @Olavhaasie are suggesting, I'd rather remove line 7 up to 27 all together. But, then I'd like to see the addition of a sole test joint launcher with input device and gait selector in the `march.launch` file. This would make the testing of gain scheduling less convoluted.

Olavhaasie: You could include the launch file for the input device and gait selector for the time being and when it has been tested we will see what we can do in `march.launch`

march_gain_scheduling/src/march_gain_scheduling/dynamic_pid_reconfigurer.py
```diff
@@ -1,39 +1,69 @@
...
+            rospy.logdebug('Config set to {0}, {1}, {2}'.format(self.current_gains[i][0],
+                                                                self.current_gains[i][1],
+                                                                self.current_gains[i][2]))
 
     # Method that pulls the PID values from the gains_per_gait_type.yaml config file
-    def look_up_table(self, i):
-        if rospy.has_param('~gait_types/' + self._gait_name + '_gains/'):
-            gains = rospy.get_param('~gait_types/' + self._gait_name + '_gains/' + self._joint_list[i])
-            return gains['p'], gains['i'], gains['d']
+    def look_up_table(self, joint_index):
```
Ishadijcks: Non-descriptive name

PR #420, opened 2019-12-04
Add slope down gait to state machine

gaiavdh:
## Description
 Add two new gaits to the state machine, both for the Ramp and Door obstacle. As the first gait (RD slope down) ends on the slope and the second gait (RD last step) starts from that position, we also created a new Idle State (because the ankle angles are different, and therefore the 'normal' standing Idle State can not be used).

## Changes
 Added two gaits to the state machine.
Added new Idle State to state machine.

### DIFF
--- a/march_state_machine/src/march_state_machine/healthy_sm.py
+++ b/march_state_machine/src/march_state_machine/healthy_sm.py
@@ -2,6 +2,7 @@ import rospy
 import smach
 from std_srvs.srv import Empty, EmptyRequest
 
+from . import ramp_door_last_step_sm
 from . import ramp_door_slope_down_sm
 from . import ramp_door_slope_up_sm
 from . import rough_terrain_high_step_sm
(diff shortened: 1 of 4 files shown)


march_state_machine/src/march_state_machine/healthy_sm.py
```diff
@@ -107,11 +109,19 @@ def create():
         smach.StateMachine.add('GAIT RD SLOPE UP', ramp_door_slope_up_sm.create(),  # RD stands for Ramp and Door
                                transitions={'succeeded': 'STANDING', 'failed': 'UNKNOWN'})
 
+        smach.StateMachine.add('GAIT RD SLOPE DOWN', ramp_door_slope_down_sm.create(),
+                               transitions={'succeeded': 'STANDING SLOPE DOWN', 'failed': 'UNKNOWN'})
+
+        smach.StateMachine.add('GAIT RD LAST STEP', ramp_door_last_step_sm.create(),
+                               transitions={'succeeded': 'STANDING', 'failed': 'UNKNOWN'})
```
Olavhaasie: I would add the `GAIT RD LAST STEP` to the `GAIT RD SLOPE DOWN` state machine. Because the last step is only useful for that state.

You should do this by adding the `STANDING SLOPE DOWN` states  and `GAIT RD LAST STEP` states to `ramp_door_slope_down_sm.py`

gaiavdh: We have done this, but a little different than you said. We have followed the example of the Tilted Path gait made by MARCH IV; that worked in the simulation.

PR #422, opened 2019-12-10
Feature: Generalize gait state machines

Olavhaasie:
Closes PM-197

## Description
This issue started with the state machine publishing on `/march/input_device/instruction_response` when a gait was finished. However, I found that the easiest way to do this was by publishing it on termination on a state machine. Then I implemented every gait as a state machine where the states are subgaits. I created two new state machines:

1. `WalkStateMachine` for gaits resembling a walking pattern
2. `StepStateMachine` for gaits that are one sequence, like step and rough terrain

This resulted in removing a lot of code :100: :fire: 

## Changes
* Added `GaitInstructionResponse.msg`
* Added state machine `WalkStateMachine` and `StepStateMachine`
* Add publish instruction response when the new state machines terminate
* Remove unused `stopped_pressed` userdata


Olavhaasie:
> In the tilted path state machine there are also a lot of gaits which are just a step. Can't these be generalized to the step state machine?

Unfortunately these gaits are implemented where every state is a new gait, so that is not possible with the current implementation. However, these gaits are used nowhere anymore so I could remove them

RutgerVanBeek:
> > In the tilted path state machine there are also a lot of gaits which are just a step. Can't these be generalized to the step state machine?
> 
> Unfortunately these gaits are implemented where every state is a new gait, so that is not possible with the current implementation. However, these gaits are used nowhere anymore so I could remove them


> > In the tilted path state machine there are also a lot of gaits which are just a step. Can't these be generalized to the step state machine?
> 
> Unfortunately these gaits are implemented where every state is a new gait, so that is not possible with the current implementation. However, these gaits are used nowhere anymore so I could remove them

But the first, middle and last step are just step, similar to a normal step. Couldn't they be implemented as a step state machine being part of the tilted path state machine?

Ishadijcks:
Can you update the name of this pr to better reflect its content?

PR #430, opened 2020-01-02
Feature/add tilted path first step straight

marnixbrands:
## Description
 Added the gaits for the first steps onto the tilted path to the state machine. There are three different versions; one uses a side step to get onto the tilted path, one to get straight onto the tilted path starting with the right foot and one straight starting with the left foot. For the side step, we made a new idle state between the first starting and second starting gait called STANDING TP SIDEWAYS. 

## Changes
 Added tilted path gaits to state machine
Made a new idle state for the side step onto the tilted path.


march_state_machine/src/march_state_machine/gaits/tilted_path_sideways_sm.py
```diff
@@ -0,0 +1,21 @@
...
+    with sm_tilted_path_sideways:
+        smach.StateMachine.add('GAIT TP STRAIGHT FIRST', StepStateMachine('tilted_path_first_start',
+                                                                          subgaits=['left_open', 'right_close']),
+                               transitions={'succeeded': 'STANDING TP SIDEWAYS', 'failed': 'failed'})
+
+        smach.StateMachine.add('STANDING TP SIDEWAYS', IdleState(outcomes=['gait_tilted_path_second_start',
+                                                                           'preempted']),
+                               transitions={'gait_tilted_path_second_start': 'GAIT TP STRAIGHT SECOND'})
+
+        smach.StateMachine.add('GAIT TP STRAIGHT SECOND', StepStateMachine('tilted_path_first_start',
```
Olavhaasie: I found the problem
```suggestion
        smach.StateMachine.add('GAIT TP STRAIGHT SECOND', StepStateMachine('tilted_path_second_start',
```

PR #432, opened 2020-01-03
Feature: possible gaits service

Olavhaasie:
Closes PM-94

## Description
This PR adds a new service server to the state machine that returns the possible next gaits. This will be used in the (developer) input device. In order to create this I had to change the healthy state machine to a class to access the current states. The new service is added to that class and is called `/march/state_machine/get_possible_gaits`. You can test in shell with this command:

    rosservice call /march/state_machine/get_possible_gaits "{}"

## Changes
* Added new service description `PossibleGaits` which returns an string array of possible gaits.
* Created `HealthyStateMachine` class
* Added `/march/state_machine/get_possible_gaits` service server


Olavhaasie:
Merge #430 first due to merge conflicts that will come

PR #433, opened 2020-01-06
Fix: remove negative difference

Olavhaasie:
Closes PM-237

## Description
There was a negative difference error when the gait was already finished before the given endtime. This would only happen during the home sit and stand gaits, since those are programmed to be finished in 3 seconds, but can sometimes be done in less. This means that the home sit and stand gaits will now sooner finish and so you can sooner continue with the next gait :tada: (although this is at most 3 seconds, which is not much).

## Changes
* Removed negative difference
* Commented unused parameters
* Removed useless `== NULL` check


march_gait_scheduler/src/gait_scheduler_node.cpp
```diff
@@ -20,7 +20,7 @@ ScheduleGaitActionServer* schedule_gait_action_server = nullptr;
 actionlib::SimpleActionClient<control_msgs::FollowJointTrajectoryAction>* follow_joint_trajectory_action;
 Scheduler* scheduler = nullptr;
 
-void doneCallback(const actionlib::SimpleClientGoalState& state,
+void doneCallback(const actionlib::SimpleClientGoalState& /* state */,
```
JorisWeeda: why not just remove the whole variable and type? or is that not possible with this callback?

march_gait_scheduler/test/CallbackCounter.cpp
```diff
@@ -12,8 +12,8 @@ struct CallbackCounter
   {
   }
 
-  void cb_action(const control_msgs::FollowJointTrajectoryGoalConstPtr& goal)
+  void cb_action(const control_msgs::FollowJointTrajectoryGoalConstPtr& /* goal */)
```
JorisWeeda: same comment as gait_scheduler_node line 23

march_gait_scheduler/src/gait_scheduler_node.cpp
```diff
@@ -101,7 +96,7 @@ void scheduleGaitCallback(const march_shared_resources::GaitGoalConstPtr& goal)
  * @param config the config file with all the parameters
  * @param level A bitmask
  */
-void schedulerConfigCallback(march_gait_scheduler::SchedulerConfig& config, uint32_t level)
+void schedulerConfigCallback(march_gait_scheduler::SchedulerConfig& config, uint32_t /* level */)
```
JorisWeeda: same comment as gait_scheduler_node line 23

march_gait_scheduler/src/gait_scheduler_node.cpp
Olavhaasie: It is not possible to remove the value otherwise the type is wrong for the action library. I commented it out to make clear that these variables were not used.

PR #436, opened 2020-01-15
Feature: sound interface

Olavhaasie:
Closes PM-158

## Description
This PR removes the `march_sound_scheduler` package and uses the ROS [`sound_play`](https://wiki.ros.org/sound_play) directly to play sounds. Since all the `march_sound_scheduler` package did was receive messages about sounds and passing them on to the `sound_play` node. `sound_play` has a good library interface for both Python and C++. The sounds are moved to the `march_safety` package which loads and plays them.

## Changes
* Removed `march_sound_scheduler`
* Created `sound` directory in `march_safety`, which contains the fatal and non fatal sounds
* Some small clang-tidy fixes while I was modifying the `march_safety` code anyway


Ishadijcks:
How will you schedule sounds to be played at a certain part of a gait?

Olavhaasie:
> How will you schedule sounds to be played at a certain part of a gait?

This will still be possible from the state machine, gait selection or gait scheduler even.

PR #438, opened 2020-01-17
Feature/pm 249 add rockerswitch input

JorisWeeda:
Closes PM-249

## Description
Added a rocker switch variable to the gaitInstruction msg that the input device sends. The variable does not have a direct purpose because the state machine has not yet been changed.

## Changes
* Added the rockerswitch instruction with corresponding variable to gaitInstruction.msg

## Tested
* On the simulation model


march_shared_resources/msg/GaitInstruction.msg
```diff
@@ -8,6 +8,10 @@ int8 STOP = 0
 int8 GAIT = 1
 int8 PAUSE = 2
 int8 CONTINUE = 3
+int8 ROCKER_SWITCH = 4
 
 # If the type is a GAIT this is the name of the gait.
 string gait_name
+
+# If the type is a ROCKER SWITCH this is would be increment or decrement
```
Ishadijcks: What are you trying to say here?

march_shared_resources/msg/GaitInstruction.msg
```diff
@@ -8,6 +8,10 @@ int8 STOP = 0
...
+int8 rocket_switch
```
Ishadijcks: typo?

march_shared_resources/msg/GaitInstruction.msg
JorisWeeda: If the rocker switch is used the given instruction will be set to ROCKER_SWITCH (4). In this case the rocker_switch variable should be set to an increment or decrement value.

march_shared_resources/msg/GaitInstruction.msg
```diff
@@ -8,6 +8,14 @@ int8 STOP = 0
...
+
+# Possible rocker switch output
+int8 INCREMENT = 1
+int8 DECREMENT = -1
 
 # If the type is a GAIT this is the name of the gait.
 string gait_name
+
+# If the instruction type is ROCKER SWITCH increment or decrement the value
+int8 rocker_switch
```
Olavhaasie: I still think the rocker switch is a bad name for this function. Since the name does not directly describe the function it is supposed to perform. The other names are `GAIT`, `STOP`, `CONTINUE` and `STOP`. Those are all actions to perform. So I see the `INCREMENT_STEP_SIZE` and `DECREASE_STEP_SIZE` as two different functions.

I would implement it like this:
```
Header header

# The type of instruction.
int8 type

# All possible instruction types.
int8 STOP = 0
int8 GAIT = 1
int8 GAIT = 1
int8 PAUSE = 2
int8 CONTINUE = 3
int8 INCREMENT_STEP_SIZE = 4
int8 DECREMENT_STEP_SIZE = 5

# If the type is a GAIT this is the name of the gait.
string gait_name
```
This implementation is more consistent with the current implementation that already exists and does not introduce one extra variable.

I also think that this naming makes it too dependent on how the current hardware is implemented. Were this ever be used by another exoskeleton, which uses a different kind of input, the `ROCKER_SWITCH` would make no sense.

Let me know what you think

JorisWeeda: When adding the rocker switch to the msg it came to mind to separate a single increment or decrement (for the variable step size) and also holding the increment button and decrement button (for possible change in speed). Therefore I created a new variable because it could span four possibilities. However, your approach seems reasonable and would better fit in the current method. So we'll go with that one.

PR #439, opened 2020-01-17
Feature/pm246 stairs in sim

Roelemans:
Closes PM-246

## Description

Stairs can be added to the simulation by adding the option " obstacle:=stairs" to march_simulation.launch


Roelemans:
That's correct - it is probably a combination of the tuning in the plugin controlling the CoM and the way the stairs gait works.

PR #440, opened 2020-01-27
Pm 238 transition step variable stepsize

JorisWeeda:
Closes PM-238

## Description
This PR implements a new class to create a transition sub-gait while the exoskeleton is executing a gait. It uses two given gait names and the sub gait name in which the transition will take place. A first version of the general calculations used to accomplish this result is explain on the confluence page; **https://confluence.projectmarch.nl:8443/display/51/Transitioning+step** . This page also has some graphs to illustrate the course of the transition in relation with the given gaits.

## Changes
* Removed the non used test
* Created a new class which generates a transition sub gait.
* Created a unit test for the transition (to test this the simulation must be running in a separate terminal)
* Added new test to the launch file 
* Added test files which has different step sizes.

## Tested
* Tested using the unit test and plotting the output of the transition class

## Notes
Dont forget to squash and merge, lot of commits to test Travis CI


Ishadijcks:
> In order to run the unit test the roslaunch march_simulation march_world.launch must be running because the gait selection nodes needs the robot information published by the simulation. If the simulation is running use colcon test to run the test and see if it passes.

Does your test not launch this dependency itself?

march_gait_selection/src/march_gait_selection/transition_gait/create_transition.py
```diff
@@ -0,0 +1,206 @@
...
+        timestamps = sorted(set(old_subgait.get_unique_timestamps() + new_subgait.get_unique_timestamps()))
+
+        for old_joint in old_subgait.joints:
+            new_joint = new_subgait.get_joint(old_joint.name)
+
+            old_joint_setpoints = []
+            new_joint_setpoints = []
+            for timestamp in timestamps:
+                old_joint_setpoints.append(old_joint.get_interpolated_setpoint(timestamp))
+                new_joint_setpoints.append(new_joint.get_interpolated_setpoint(timestamp))
```
Roelemans: You add a setpoint even for the joints that already have a setpoint there. Every joint will probably have two setpoints at zero now.

march_gait_selection/src/march_gait_selection/transition_gait/create_transition.py
```diff
@@ -0,0 +1,206 @@
...
+                new_joint = new_subgait.get_joint(old_joint.name)
+
+                if len(old_joint) != len(new_joint):
+                    rospy.logerr('Amount of setpoints is inconsistent between subgait {og} of gait {fg} and {sg}'
+                                 .format(og=old_subgait.subgait_name, fg=old_subgait.gait_name,
+                                         sg=new_subgait.gait_name))
+
+                joints.append(new_joint.name)
+
+                new_setpoint_factor = 1.0 / (total_transition_factor - 1) * transition_index
```
Roelemans:
```suggestion
                new_setpoint_factor = transition_index / (total_transition_factor - 1.0)
```
Avoids ambiguity in order of computation

march_gait_selection/src/march_gait_selection/transition_gait/create_transition.py
```diff
@@ -0,0 +1,206 @@
...
+        :returns:
+             list of joints, velocities and positions corresponding to every joint at given timestamps and duration
+        """
+        joints = []
+        transition_positions = []
+        transition_velocities = []
+        transition_durations = []
+
+        total_transition_factor = len(new_subgait[0].setpoints)
+        for transition_index in range(total_transition_factor):
```
Roelemans:
```suggestion
        for setpoint_index in range(number_setpoints):
```
I think it will be more intuitive if you name them after what they are rather than what they are used for

march_gait_selection/src/march_gait_selection/transition_gait/create_transition.py
```diff
@@ -0,0 +1,206 @@
...
+
+            positions = []
+            velocities = []
+            for old_joint in old_subgait.joints:
+                new_joint = new_subgait.get_joint(old_joint.name)
+
+                if len(old_joint) != len(new_joint):
+                    rospy.logerr('Amount of setpoints is inconsistent between subgait {og} of gait {fg} and {sg}'
+                                 .format(og=old_subgait.subgait_name, fg=old_subgait.gait_name,
+                                         sg=new_subgait.gait_name))
```
Roelemans: Your current program will error if len(old_joint) or len(new_joint) are not equal to "total_transition_factor". You should check for that instead. (if you change the order of loops here as suggested, the current check is the correct one of course)

march_gait_selection/src/march_gait_selection/transition_gait/create_transition.py
```diff
@@ -0,0 +1,206 @@
...
+                                 + (new_setpoint.position * new_setpoint_factor))
+                velocities.append((old_setpoint.velocity * old_setpoint_factor)
+                                  + (new_setpoint.velocity * new_setpoint_factor))
+
+                trans_pos = (old_setpoint.position * old_setpoint_factor) + \
+                            (new_setpoint.position * new_setpoint_factor)
+
+                rospy.logwarn('old factor: {of} \t new factor {nf} \t old: {op} \t transition: {tp} \t new: {np} '
+                              .format(of=str(old_setpoint_factor), nf=str(new_setpoint_factor),
+                                      op=str(old_setpoint.position), tp=str(trans_pos), np=str(new_setpoint.position)))
```
Roelemans: Why this warning? Why only for position not for velocity? You should just use positions[-1] instead of recalculating with trans_pos, reduces the duplicate code and makes this less error prone :)
```suggestion
                                      op=str(old_setpoint.position), tp=str(positions[-1]), np=str(new_setpoint.position)))
```

march_gait_selection/src/march_gait_selection/transition_gait/create_transition.py
```diff
@@ -0,0 +1,206 @@
...
+        return old_subgait, new_subgait
+
+    @staticmethod
+    def _transition_trajectory(old_subgait, new_subgait):
+        """Calculate a transition trajectory which starts at the old gait and ends with the endpoints of the new gait.
+
+        :param old_subgait:
+            The subgait which the transition is starting from
+        :param new_subgait:
+            The subgait which the transition will result in.
```
Roelemans:
```suggestion
            The subgait which the transition will result in
```

march_gait_selection/src/march_gait_selection/transition_gait/create_transition.py
```diff
@@ -0,0 +1,206 @@
...
+                                      op=str(old_setpoint.position), tp=str(trans_pos), np=str(new_setpoint.position)))
+
+            transition_positions.append(positions)
+            transition_velocities.append(velocities)
+            transition_durations.append(new_setpoint.time)
+
+        return joints, transition_positions, transition_velocities, transition_durations
+
+    @staticmethod
+    def _to_subgait_dictionary(joints, positions_list, velocity_list, durations):
```
Roelemans: Would it not be more efficient to build JointTrajectory objects during  _transtition_trajectroy() and then just use the __init__ of Subgait? Seems a bit roundabout to transform this into the somewhat strange ros message format in order to use from_dict().

march_shared_classes/src/march_shared_classes/exceptions/gait_exceptions.py
```diff
@@ -48,3 +48,29 @@ def __init__(self, gait_name=None, msg=None):
             msg = 'The given gait: {gn} has incorrect information'.format(gn=gait_name)
 
         super(NonValidGaitContent, self).__init__(msg)
+
+
+class NonMarchingGaitContent(GaitError):
```
Roelemans:
```suggestion
class NonMatchingGaitContent(GaitError):
```

march_gait_selection/src/march_gait_selection/transition_gait/create_transition.py
```diff
@@ -0,0 +1,206 @@
...
+from march_shared_classes.exceptions.gait_exceptions import GaitError, SubgaitNameNotFound
+from march_shared_classes.gait.subgait import Subgait
+
+
+SUBGAIT_NAME = 'transition_subgait'
+GAIT_NAME = 'transition'
+VERSION = 'default'
+
+
+class TransitionSubgait(object):
```
Roelemans: For a beautiful integration with the GaitSelection object you might want to consider something like the following:

Have this class inherit from Subgait and create a Gait that only contains this Subgait. When the Gait is then called by PerformGaitAction, add Subgaits to the Gait object that belong to one of the walking gaits and generate the relevant content for the TransitionSubgait object (i.e. JointTrajectories).

I did not completely think this out, but maybe something like this is possible.

march_gait_selection/src/march_gait_selection/transition_gait/create_transition.py
JorisWeeda: This warning is deprecated, I was using it to get some information in the logfiles

march_gait_selection/src/march_gait_selection/transition_gait/transition_subgait.py
```diff
@@ -0,0 +1,172 @@
...
+    @staticmethod
+    def _get_limits(robot, joint_name):
+        """Use the parsed robot to get the defined joint limits."""
+        for urdf_joint in robot.joints:
+            if urdf_joint.name == joint_name:
+                return Limits(urdf_joint.safety_controller.soft_lower_limit,
+                              urdf_joint.safety_controller.soft_upper_limit,
+                              urdf_joint.limit.velocity)
+
+        raise TransitionError('No robot selected for create the transition gait')
```
Roelemans:
```suggestion
        raise TransitionError('robot does not contain joint {joint}'.format(joint=joint_name))
```

march_gait_selection/src/march_gait_selection/transition_gait/transition_subgait.py
```diff
@@ -0,0 +1,172 @@
...
+
+    @staticmethod
+    def _scale_timestamps_subgaits(old_subgait, new_subgait):
+        """Scale all the setpoint to match the duration in both subgaits."""
+        old_duration = old_subgait.duration
+        new_duration = new_subgait.duration
+
+        for joint in old_subgait.joints:
+            for setpoint in joint.setpoints:
+                setpoint.time = setpoint.time / old_duration * new_duration
```
Roelemans:
```suggestion
                setpoint.time = setpoint.time * new_duration / old_duration 
```

march_gait_selection/src/march_gait_selection/transition_gait/transition_subgait.py
```diff
@@ -0,0 +1,172 @@
...
+        transition_joints = cls._transition_joints(gait_selection.robot, old_subgait, new_subgait)
+        transition_duration = new_subgait.duration
+
+        transition_subgait = cls(transition_joints, transition_duration)
+        cls._validate_transition_gait(old_subgait, transition_subgait, new_subgait)
+
+        return transition_subgait
+
+    @staticmethod
+    def _get_subgaits(gait_selection, old_gait_name, new_gait_name, new_subgait_name, old_subgait_name):
```
Roelemans:
```suggestion
    def _get_copy_of_subgaits(gait_selection, old_gait_name, new_gait_name, new_subgait_name, old_subgait_name):
```

march_gait_selection/src/march_gait_selection/transition_gait/create_transition.py
Roelemans: Never mind this is not true.

JorisWeeda:
comments for roel

march_gait_selection/test/nosetests/test_gait_transition.py
```diff
@@ -0,0 +1,55 @@
...
+before_subgait_name = 'left_swing'
+subgait_name = 'right_swing'
+after_subgait_name = 'left_close'
+
+
+class TestTransitionTrajectory(unittest.TestCase):
+    def setUp(self):
+        self.gait_selection = GaitSelection(PKG, DIR)
+
+    def test_valid_gaits(self):
```
Roelemans: Also make a test_invalid_gaits tot test if some random gait name actually returns None (otherwise this test breaks when someone decides to return something else than None when a gait does not exist)

march_gait_selection/test/nosetests/test_gait_transition.py
```diff
@@ -0,0 +1,55 @@
...
+        last_gait = False if self.gait_selection[old_gait_name] is None else True
+        new_gait = False if self.gait_selection[new_gait_name] is None else True
+
+        self.assertTrue(last_gait, msg='{gn} gait could not be found in map'.format(gn=old_gait_name))
```
Roelemans:
```suggestion
        self.assertTrue(last_gait, msg='{gn} gait could not be found in folder'.format(gn=old_gait_name))
```

march_gait_selection/test/nosetests/test_gait_transition.py
```diff
@@ -0,0 +1,55 @@
...
+        self.assertTrue(new_gait, msg='{gn} gait could not be found in map'.format(gn=new_gait_name))
```
Roelemans:
```suggestion
        self.assertTrue(new_gait, msg='{gn} gait could not be found in fodler'.format(gn=new_gait_name))
```

march_gait_selection/src/march_gait_selection/transition_gait/transition_subgait.py
```diff
@@ -0,0 +1,172 @@
...
+                old_joint_setpoints.append(old_joint.get_interpolated_setpoint(timestamp))
+                new_joint_setpoints.append(new_joint.get_interpolated_setpoint(timestamp))
+
+            old_joint.setpoints = old_joint_setpoints
+            new_joint.setpoints = new_joint_setpoints
+
+        return old_subgait, new_subgait
+
+    @staticmethod
+    def _scale_timestamps_subgaits(old_subgait, new_subgait):
```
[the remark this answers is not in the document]
JorisWeeda: why use the average? it is only necessary to change the old sub gait duration and the new gait remains unchanged.

march_gait_selection/test/nosetests/test_gait_transition.py
```diff
@@ -0,0 +1,55 @@
...
+after_subgait_name = 'left_close'
+
+
+class TestTransitionTrajectory(unittest.TestCase):
+    def setUp(self):
+        self.gait_selection = GaitSelection(PKG, DIR)
+
+    def test_valid_gaits(self):
+        last_gait = False if self.gait_selection[old_gait_name] is None else True
+        new_gait = False if self.gait_selection[new_gait_name] is None else True
```
Roelemans: You can just split these tests

march_shared_classes/package.xml
```diff
@@ -11,4 +11,5 @@
   <depend>march_shared_resources</depend>
 
   <exec_depend>rospy</exec_depend>
+  <exec_depend>python-scipy</exec_depend>
```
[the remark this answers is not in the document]
JorisWeeda: Yes, the colcon test module reported a missing module when executing the test. Previously these methods were only build by Travis CI and (maybe) tested locally on a MARCH laptop which already has this module.

march_gait_selection/test/nosetests/test_gait_transition.py
```diff
@@ -0,0 +1,55 @@
...
+        self.assertTrue(after_subgait, msg='{sg} subgait could not be found in gait {gn}'
+                        .format(sg=after_subgait_name, gn=new_gait_name))
+
+        subgait_small = False if self.gait_selection[old_gait_name][subgait_name] is None else True
+        subgait_medium = False if self.gait_selection[new_gait_name][subgait_name] is None else True
+
+        self.assertTrue(subgait_small, msg='{sg} subgait could not be found in gait {gn}'
+                        .format(sg=subgait_name, gn=old_gait_name))
+        self.assertTrue(subgait_medium, msg='{sg} subgait could not be found in gait {gn}'
+                        .format(sg=subgait_name, gn=new_gait_name))
```
Roelemans: Split these into four seperate tests too #moretestsmorebetter

march_gait_selection/test/nosetests/test_gait_transition.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def test_valid_transition_medium_to_small(self):
+        TransitionSubgait.from_subgait_names(self.gait_selection, old_gait_name, new_gait_name, subgait_name)
```
Roelemans: I guess you are testing that this raises no error? Maybe nice to place a comment here

march_gait_selection/test/nosetests/test_gait_transition.py
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env python
```
Roelemans: In principle, you will want to do more unit-testing, i.e. test smaller parts of the code with a single test. Every method should have at least one dedicated test (and usually more to cover all branches/different types of input). This way, you will be able to find problems much more quickly (because you know exactly which method errors).

march_gait_selection/test/nosetests/test_gait_transition.py
JorisWeeda: I agree that this should be tested. However, testing wrong gait and subgait names seems more like a unittest for the gait selection module. I will create this in a separate PR. Testing the existing names of walk small, walk medium, right swing, left swing, left close are only executed because otherwise the transition subgait might fail.

march_gait_selection/src/march_gait_selection/transition_gait/transition_subgait.py
```diff
@@ -0,0 +1,172 @@
...
+            raise SubgaitNameNotFound(subgait_name=new_subgait_name)
+
+        return old_subgait, new_subgait
+
+    @staticmethod
+    def _equalize_amount_of_setpoints(old_subgait, new_subgait):
+        """Equalize the subgaits to have matching amount of setpoints on all the timestamps."""
+        timestamps = sorted(set(old_subgait.get_unique_timestamps() + new_subgait.get_unique_timestamps()))
+
+        for old_joint in old_subgait.joints:
```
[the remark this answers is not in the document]
JorisWeeda: I agree that this function from the modifiable joint trajectory should be moved to the normal joint trajectory. But, it seems to me like something for a new PR since the gait generator will also be changed. Maybe a bit overkill for this PR only.

march_gait_selection/src/march_gait_selection/transition_gait/transition_subgait.py
```diff
@@ -0,0 +1,172 @@
...
+            number_setpoints = len(new_subgait[0].setpoints)
+            for transition_index in range(number_setpoints):
+                new_setpoint_factor = transition_index / (number_setpoints - 1.0)
+                old_setpoint_factor = 1.0 - new_setpoint_factor
+
+                old_setpoint = old_joint[transition_index]
+                new_setpoint = new_joint[transition_index]
+
+                position = (old_setpoint.position * old_setpoint_factor) + (new_setpoint.position * new_setpoint_factor)
+                velocity = (old_setpoint.velocity * old_setpoint_factor) + (new_setpoint.velocity * new_setpoint_factor)
```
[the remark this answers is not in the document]
JorisWeeda: Not sure what u are implying here. As of now creating the set point has been migrated to a separate function to split the functionalities. Maybe that is what u mean.

march_gait_selection/test/nosetests/test_gait_transition.py
JorisWeeda: Indeed, the existing test function just check the general output. Now that the structure is relatively solid I will add more testing functions.

march_gait_selection/src/march_gait_selection/transition_gait/transition_subgait.py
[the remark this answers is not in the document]
Roelemans: Very good! :+1:

march_gait_selection/test/nosetests/test_gait_transition.py
```diff
@@ -21,71 +21,70 @@
 
 wrong_name = 'wrong'
 
+gait_selection = GaitSelection(PKG, DIR)
```
Roelemans: Why was this necessary?

march_gait_selection/test/nosetests/test_gait_transition.py
Roelemans: I think you didn't do this?

march_gait_selection/src/march_gait_selection/transition_gait/transition_subgait.py
```diff
@@ -159,13 +158,53 @@ def _get_limits(robot, joint_name):
...
+        for transition_joint in transition_subgait.joints:
+            old_joint = old_subgait.get_joint(transition_joint.name)
+            new_joint = new_subgait.get_joint(transition_joint.name)
+
+            for old_setpoint, transition_setpoint, new_setpoint in zip(old_joint, transition_joint, new_joint):
+                if old_setpoint.time != transition_setpoint.time or new_setpoint.time != transition_setpoint.time:
+                    raise TransitionError('The transition timestamps are not equal to the old and new subgait')
+
+                if old_setpoint.position < transition_setpoint.position:
+                    if not transition_setpoint.position <= new_setpoint.position:
```
Roelemans: Shouldn't you also check that it is above old_setpoint.position? I think if old_postion = 0, new_position = 1 and transition_position = -100000, then this method will accept it as valid. (Same for the code below)

march_gait_selection/src/march_gait_selection/transition_gait/transition_subgait.py
```diff
@@ -104,16 +104,15 @@ def _equalize_amount_of_setpoints(old_subgait, new_subgait):
...
         old_duration = old_subgait.duration
-        new_duration = new_subgait.duration
 
         for joint in old_subgait.joints:
             for setpoint in joint.setpoints:
-                setpoint.time = setpoint.time / old_duration * new_duration
+                setpoint.time = setpoint.time * new_duration / old_duration
 
-        return old_subgait, new_subgait
+        return old_subgait
```
Roelemans: You don't really need to return anything here, since you directly edit old_subgait anyway. It can't do harm either though, and this might make the code a bit clearer. Do what you prefer.

march_gait_selection/test/nosetests/test_gait_transition.py
JorisWeeda: Because leaving it in the setup of the unittest would mean the parameterServer would be called every test case to get the xacro data, that would increase the testing time dramatically. Constructing the object outside the testCase would be much more efficient.

march_gait_selection/src/march_gait_selection/transition_gait/transition_subgait.py
```diff
@@ -159,13 +158,53 @@ def _get_limits(robot, joint_name):
...
+        position = (old_setpoint.position * old_factor) + (new_setpoint.position * new_factor)
+        velocity = (old_setpoint.velocity * old_factor) + (new_setpoint.velocity * new_factor)
+
+        return Setpoint(new_setpoint.time, position, velocity)
+
+    @staticmethod
+    def _get_all_unique_timestamps(old_subgait, new_subgait):
+        """Get all the timestamps from the subgaits, eliminate double."""
+        timestamps = sorted(set(old_subgait.get_unique_timestamps() + new_subgait.get_unique_timestamps()))
+        return [round(timestamp, Setpoint.digits) for timestamp in timestamps]
```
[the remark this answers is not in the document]
JorisWeeda: The computer has some internal rounding error which causes a very (very) small change in timestamps. Therefore the rounding is necessary. I intentionally did not change the equal to testing condition because if we rounded it in the first place we should keep this standard.

march_gait_selection/src/march_gait_selection/transition_gait/transition_subgait.py
```diff
@@ -84,17 +87,28 @@ def _get_copy_of_subgaits(gait_selection, old_gait_name, new_gait_name, new_subg
...
+            for setpoint in joint.setpoints:
+                setpoint.time = setpoint.time * new_duration / old_duration
+
+        return old_subgait
+
     @staticmethod
     def _equalize_amount_of_setpoints(old_subgait, new_subgait):
         """Equalize the subgaits to have matching amount of setpoints on all the timestamps."""
-        timestamps = sorted(set(old_subgait.get_unique_timestamps() + new_subgait.get_unique_timestamps()))
+        unique_timestamps = TransitionSubgait._get_all_unique_timestamps(old_subgait, new_subgait)
```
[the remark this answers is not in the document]
JorisWeeda: It is in a separate method because in the future this function may be used for other purposes.

march_gait_selection/src/march_gait_selection/transition_gait/transition_subgait.py
JorisWeeda: I don't think this is true. if the transition_position is -100000 and old position is 0. Then the second if statement is called;
 ```suggestion
                if old_setpoint.position > transition_setpoint.position:
                    if not transition_setpoint.position >= new_setpoint.position:
                        raise TransitionError
```
If in the scenario u are describing the new setpoint is 1 then the transition error is raised because it is not between the two given setpoints.

The method is arranged as followed:
The transition position is always **between** the two given setpoints because otherwise the transition would not fit the pattern of the setpoints. Earlier in the code the startpoint and endpoint are validated using the **validate trajectory** in the gait class. This function purely validates if the setpoints is between the old-new or new-old.

File: march_gait_selection/test/nosetests/test_gait_transition.py (hunk 24 above)
Comment (Roelemans) [reply march_gait_selection/test/nosetests/test_gait_transition.py]: Downside of this might be that if gait_selection is modified in one test, then that modification will exist in all subsequent tests. But as long as you pay attention to that I guess it is fine :+1:

File: march_gait_selection/src/march_gait_selection/transition_gait/transition_subgait.py (hunk 25 above)
Comment (Roelemans) [reply march_gait_selection/src/march_gait_selection/transition_gait/transition_subgait.py]: Ah never mind my brain errored for a moment there

Comment (Roelemans) [inline march_gait_selection/src/march_gait_selection/transition_gait/transition_subgait.py]:
```suggestion
                    if new_setpoint.position < transition_setpoint.position:
```
Don't use an unnecessary amount of negations, and keep order of inequalities consistent. This makes things a bit easier to follow :)

march_gait_selection/src/march_gait_selection/transition_gait/transition_subgait.py
```diff
@@ -159,13 +158,53 @@ def _get_limits(robot, joint_name):
...
+                    raise TransitionError('The transition timestamps are not equal to the old and new subgait')
+
+                if old_setpoint.position < transition_setpoint.position:
+                    if not transition_setpoint.position <= new_setpoint.position:
+                        raise TransitionError('The transition position {tp} is not between the old {op} and new {np}'
+                                              .format(tp=transition_setpoint.position, op=old_setpoint.position,
+                                                      np=new_setpoint.position))
+
+                if old_setpoint.position > transition_setpoint.position:
+                    if not transition_setpoint.position >= new_setpoint.position:
```
Roelemans:
```suggestion
                    if new_setpoint.position > transition_setpoint.position:
```

march_gait_selection/test/nosetests/test_gait_selection.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def test_right_close_walk_medium(self):
+        # check if subgait name exists
+        is_right_close_subgait = False if gait_selection[walk_medium][right_close] is None else True
+        self.assertTrue(is_right_close_subgait, msg='{sg} subgait could not be found in gait {gn}'
+                        .format(sg=right_close, gn=walk_medium))
+
+
+if __name__ == '__main__':
+    rosunit.unitrun('march_gait_selection', 'gait_selection_test', TestGaitSelection)
```
Olavhaasie: This addition is not necessary as this file is never executed on its own. It is executed through `run_tests`

march_gait_selection/test/nosetests/test_gait_transition.py
```diff
@@ -0,0 +1,74 @@
...
+        #  Test if the TransitionSubgait is created without an error
+        TransitionSubgait.from_subgait_names(gait_selection, walk_small, walk_medium, right_close)
+
+    def test_valid_transition_medium_to_small_right_close(self):
+        #  Test if the TransitionSubgait is created without an error
+        TransitionSubgait.from_subgait_names(gait_selection, walk_medium, walk_small, right_close)
+
+
+if __name__ == '__main__':
+    rosunit.unitrun('march_gait_selection', 'transition_gait_test', TestTransitionTrajectory)
```
Olavhaasie: Same here

march_gait_selection/CMakeLists.txt
```diff
@@ -2,31 +2,30 @@ cmake_minimum_required(VERSION 2.8.3)
...
-    CATKIN_DEPENDS
+catkin_package(CATKIN_DEPENDS
+    march_description
     march_shared_classes
     march_shared_resources
     std_msgs
 )
 
-catkin_install_python(PROGRAMS scripts/${PROJECT_NAME}_node
+catkin_install_python(PROGRAMS scripts/${PROJECT_NAME}_node test/run_tests
```
Olavhaasie: It is not necessary to install tests, at least it isn't in other packages
```suggestion
catkin_install_python(PROGRAMS scripts/${PROJECT_NAME}_node
```

march_gait_selection/CMakeLists.txt
```diff
@@ -2,31 +2,30 @@ cmake_minimum_required(VERSION 2.8.3)
...
     DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
 )
-install(DIRECTORY launch
+
+install(DIRECTORY launch test/launch
```
Olavhaasie: It is not necessary to install these launch files and also you do not want users accidentally launching tests instead of the node
```suggestion
install(DIRECTORY launch
```

march_gait_selection/CMakeLists.txt
```diff
@@ -19,14 +20,8 @@ catkin_package(
 catkin_install_python(PROGRAMS scripts/${PROJECT_NAME}_node
     DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
 )
-install(DIRECTORY launch
-    DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}
-)
```
Olavhaasie: This is still necessary :sweat_smile:

PR #441, opened 2020-01-28
Feature/add side step off tilted path to state machine

marnixbrands:
## Description

I added the ending steps off the tilted path (sideways) to the state machine. Therefore, I created a new state machine called `tilted_path_sideways_end_sm`. This because after the first ending step you end not in the STANDING IDLE state, but in the IDLE state TP SIDEWAYS END, so only the second ending step can be executed from that state. Furthermore, I changed the name of `tilted_path_sideways_sm` into `tilted_path_sideways_start_sm` to make clear distinction between stepping on or off the tilted path.
## Changes

New state machine:
`tilted_path_sideways_end_sm`

Renamed state machine:
`tilted_path_sideways_start_sm`

Added ending steps to `healthy_sm`

### DIFF
--- a/march_state_machine/src/march_state_machine/healthy_sm.py
+++ b/march_state_machine/src/march_state_machine/healthy_sm.py
@@ -5,7 +5,8 @@ from std_srvs.srv import Empty, EmptyRequest
 from march_shared_resources.srv import PossibleGaits
 
 from .gaits import ramp_down_sm
-from .gaits import tilted_path_sideways_sm
+from .gaits import tilted_path_sideways_end_sm
+from .gaits import tilted_path_sideways_start_sm
 from .state_machines.slope_state_machine import SlopeStateMachine
 from .state_machines.step_state_machine import StepStateMachine
 from .state_machines.walk_state_machine import WalkStateMachine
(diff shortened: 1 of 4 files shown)


march_state_machine/src/march_state_machine/healthy_sm.py
```diff
@@ -118,7 +121,8 @@ def __init__(self):
                               'gait_ramp_door_slope_down': 'GAIT RD RAMP DOWN',
                               'gait_tilted_path_straight_start_right': 'GAIT TP STRAIGHT START RIGHT',
                               'gait_tilted_path_straight_start_left': 'GAIT TP STRAIGHT START LEFT',
-                              'gait_tilted_path_first_start': 'GAIT TP SIDEWAYS'})
+                              'gait_tilted_path_first_start': 'GAIT TP SIDEWAYS START',
+                              'gait_tilted_path_first_end': 'GAIT TP SIDEWAYS END'})
```
Olavhaasie: I would call these transitions the same as the states without the `first`
```python
                              'gait_tilted_path_start': 'GAIT TP SIDEWAYS START',
                              'gait_tilted_path_end': 'GAIT TP SIDEWAYS END'
```

marnixbrands: The gaits are called `tilted_path_first_start` and `tilted_path_first_end`, and you have to call the transition like `gait_gait name` right? Or I can change the name of the IDLE state to GAIT TP SIDEWAYS FIRST START.

Olavhaasie: Oh now I see how they are. The gait is called first for the first step and start and end are for the beginning and the end of the obstacle

PR #445, opened 2020-01-31
Feature/test functions gait classes

Roelemans:
Closes PM-

## Description

Wrote tests for the Limits, Setpoint and JointTrajectory classes, and made some changes to the code for better testing.

## Changes

* Added tests for Limits, Setpoint and JointTrajectory
* Added method to JointTrajectory: _validate_boundary_points that checks the boundary setpoints for problems
* Moved method interpolate_setpoints
* Wrote __eq__ and __ne__ for the Setpoint class


march_shared_classes/src/march_shared_classes/gait/joint_trajectory.py
```diff
@@ -62,6 +62,8 @@ def validate_joint_transition(self, joint):
         :returns:
             True if ending and starting point are identical else False
         """
+        self._validate_boundary_points()
```
Olavhaasie: This method returns a `bool` but this value is not used

Roelemans: Oops :sweat_smile: fixed!

march_shared_classes/test/gait_test.py
```diff
@@ -0,0 +1,8 @@
+import unittest
+
+# from march_shared_classes.gait.gait import Gait
+
+
+class GaitTest(unittest.TestCase):
+    def test_something(self):
+        self.assertEqual(True, True)
```
JorisWeeda: why?

march_shared_classes/test/subgait_test.py
```diff
@@ -0,0 +1,9 @@
+import unittest
+
+# from march_shared_classes.gait.subgait import Subgait
+# from march_shared_classes.gait.joint_trajectory import JointTrajectory
+
+
+class SubgaitTest(unittest.TestCase):
+    def test_something(self):
+        self.assertEqual(True, True)
```
JorisWeeda: again, why?

march_shared_classes/test/limits_test.py
```diff
@@ -0,0 +1,17 @@
...
+        self.limits = Limits(0, 1, 2)
+
+    def test_lower_limit(self):
+        self.assertEqual(self.limits.lower, 0, 'Lower limit not initialised correctly.')
+
+    def test_upper_limit(self):
+        self.assertEqual(self.limits.upper, 1, 'Upper limit not initialised correctly.')
+
+    def test_velocity_limit(self):
+        self.assertEqual(self.limits.velocity, 2, 'Velocity limit not initialised correctly.')
```
JorisWeeda: I get the point of unit testing, but this seems a bit too much. If the limit is not constructed properly then something would be wrong with the python interpreter (which would be very unlikely). And if so, a lot of other errors would occur.

march_shared_classes/test/gait_test.py
Roelemans: Small setup for tests of the Gait class. Not currently included in the actual test paths (see run_tests.py) will be worked out properly in the next PR. Guess it was neater not to add this in this PR, but I already had it an removing it was too much effort.

march_shared_classes/test/subgait_test.py
Roelemans: Again _instert blaatverhaal here_

march_shared_classes/test/limits_test.py
Roelemans: Main reason this exists is because I started writing tests for the Limits class because it was simple. I guess these tests can still be somewhat useful - if someone decides to rename velocity to speed, or interchanges the order of the parameters in the constructor,  then this test will warn them that they probably broke shit. 
So I think we might as well leave it now that it exists.

JorisWeeda: That is true. U could keep them just to be safe. But if someone decides to edit the naming only in the class and not in the whole project he/she would be very foolish!!

march_shared_classes/test/gait_test.py
JorisWeeda: Oh sure. ok!

march_shared_classes/test/limits_test.py
Roelemans: Life would be boring without such people

PR #447, opened 2020-02-04
Add three new separate middle steps to state machine.

gaiavdh:
## Description
 We have made three separate gaits for the middle steps of the Rough Terrain, as an alternative to the middle steps gait that we used before. We saw in the training that that middle steps gait was difficult for Sjaan, so we made three separate steps that Sjaan can select herself whenever she is ready to take the next step. These three gaits are called, unsurprisingly, 'first middle step, 'second middle step' and 'third middle step'. I added these three steps to the state machine. (I did leave the 'old' middle steps gait in the state machine, so we can still use it if needed.)

## Changes
 Added three new gaits to the state machine.

### DIFF
--- a/march_state_machine/src/march_state_machine/healthy_sm.py
+++ b/march_state_machine/src/march_state_machine/healthy_sm.py
@@ -75,6 +75,11 @@ class HealthyStateMachine(smach.StateMachine):
                                                                  'right_swing', 'left_close']),
                        'STANDING')
 
+        # These are three separate steps for the middle steps of the RT, as an alternative to the middle steps above.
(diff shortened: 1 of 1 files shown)


Roelemans:
Currently you go to "standing" after every of you subgaits which means that you could select another gait after the first step (for example, walk medium). Wouldn't it be better to only be able to transfer to second middle step after completing the first?

gaiavdh:
I do agree with you that it is a little safer, but I do not think it will be the best idea at this point. We are still testing a lot, so we do not know if these steps are exactly the right size and if they work. Like we saw in the last training, maybe a different sequence of steps will eventually work better (e.g. high step, small single step, medium single step). I think it is very useful if we are able to select another gait instead of one of the three gaits at this point. But maybe later, when we have made a decision on the steps, we can add this!

PR #449, opened 2020-02-06
Added more tests to verify the transition_subgait.py

JorisWeeda:
Added more tests to verify the transition subgait module

## Description
The scaling and equalizing method of the transition_subgait.py are now unit tested as well. Also the stairs_up.gait has been added to the testing files to test that even very difficult subgaits can be used in the transition module.

## Changes
* Updated the unit tests of the transition gait module
* Added stairs_up.gait (with subgait files) for testing purposes
* Updated the transition_subgait.py


march_gait_selection/test/nosetests/test_gait_transition.py
```diff
@@ -43,26 +45,82 @@ def test_invalid_gait_selection(self):
         with self.assertRaises(GaitError):
             TransitionSubgait.from_subgait_names(wrong_name, walk_small, walk_medium, right_swing)
 
-    def test_valid_transition_small_to_medium_right_open(self):
+    def test_scale_setpoints_when_new_duration_is_larger(self):
+        # check if the scaling of the setpoints works with a large scaling factor
+        subgait_walk_medium = deepcopy(gait_selection[walk_medium][right_swing])
```
Roelemans: You could do a deep copy of the gait selection object in your setup so that you don't need to do these deepcopies in the tests.

march_gait_selection/test/nosetests/test_gait_transition.py
```diff
@@ -43,26 +45,82 @@ def test_invalid_gait_selection(self):
...
+        scaling_factor = 45.453
+
+        scaled_subgait_walk_medium = TransitionSubgait._scale_timestamps_subgaits(subgait_walk_medium, scaling_factor)
+
+        for joint in scaled_subgait_walk_medium.joints:
+            self.assertEqual(joint.setpoints[-1].time, scaling_factor,
+                             msg='Scaled end point {ep} does not match scaling factor {sf} after scaling'
+                             .format(ep=joint.setpoints[-1].time, sf=scaling_factor))
```
Roelemans: I believe you can use assertListEqual to test for the whole list at once.

march_gait_selection/test/nosetests/test_gait_transition.py
```diff
@@ -43,26 +45,82 @@ def test_invalid_gait_selection(self):
...
+
+    def test_equalize_amount_of_setpoints_with_lower_duration_new_gait(self):
+        # test if interpolation works with higher duration and different setpoints
+        subgait_walk_medium = deepcopy(gait_selection[walk_medium][right_swing])
+        subgait_stairs_up = deepcopy(gait_selection[stairs_up][right_swing])
+
+        subgait_walk_medium = TransitionSubgait._scale_timestamps_subgaits(subgait_stairs_up,
+                                                                           subgait_walk_medium.duration)
+
+        TransitionSubgait._equalize_amount_of_setpoints(subgait_stairs_up, subgait_walk_medium)
```
Roelemans: Shouldn't you assert something here? for example that the length of the subgaits is indeed equal and perhaps that the _get_unique_setpoints returns the same list for both now.

JorisWeeda: brainfart

march_gait_selection/test/nosetests/test_gait_transition.py
JorisWeeda: not entirely. However, there was a more suitable approach using the all() function in python.

march_gait_selection/src/march_gait_selection/transition_gait/transition_subgait.py
```diff
@@ -119,9 +119,14 @@ def _equalize_amount_of_setpoints(old_subgait, new_subgait):
...
+
             for timestamp in unique_timestamps:
-                old_joint_setpoints.append(old_joint.get_interpolated_setpoint(timestamp))
-                new_joint_setpoints.append(new_joint.get_interpolated_setpoint(timestamp))
+                interpolated_setpoint_old_joint = old_joint.get_interpolated_setpoint(timestamp)
+                interpolated_setpoint_new_joint = new_joint.get_interpolated_setpoint(timestamp)
+
+                if interpolated_setpoint_old_joint.time != 0 and interpolated_setpoint_new_joint.time != 0:
+                    old_joint_setpoints.append(interpolated_setpoint_old_joint)
+                    new_joint_setpoints.append(interpolated_setpoint_new_joint)
```
Roelemans: You also exclude the first setpoint that should be zero. Also you should fix the root of the issue, cause you will be missing some setpoints this way.

JorisWeeda: u were absolutely right.

march_gait_selection/test/nosetests/test_gait_transition.py
```diff
@@ -23,46 +24,104 @@
 
 
 class TestTransitionTrajectory(unittest.TestCase):
+    def setUp(self):
+        self.gait_selection = deepcopy(gait_selection)
```
Olavhaasie: Instead of a global variable you can create the `GaitSelection` object in the [`setUpClass`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUpClass) of the the test class. The `setUp` will still be called every time before a test, whereas the `setUpClass` will be called once to instantiate the `TestCase`

JorisWeeda: I did this in a previous version. However the problem there is that for every test the constructor of the gait selection module is called and in this init the parameter server is read for some data. This takes a lot of time which is not very efficient. Using a deepcopy would be a lot more efficient.

Olavhaasie: You can still use the `deepcopy` in the `setUp`, however I think it is nicer to create the `gait_selection` not globally but _once_ when the class is created.

JorisWeeda: ahh I see, would be nicer indeed

PR #452, opened 2020-02-12
Feature: pressure sole connection

RutgerVanBeek:
Closes PM-186

## Description

This PR includes the connection for the pressure soles to the software and through ESP. 
The pressure soles should be connected with an Android phone with the Moticon app, which sends the data to a windows computer, which sends it trough a socket to the exoskeleton. The data in the socket is received in the data collector. It is sends over a ROS message to the ESP adapter, which sends it to ESP. Furthermore the data collector also sends the joint states to the Moticon software. In the Moticon software more gait specific options for analysis are possible, the data about the joint states can be used then. 
## Changes

-Changed the data collector to send data to and receive data from the Moticon desktop software
-Added a pressure sole msg and topich /march/pressue_soles
-Changed the ESP adapter to also send the pressure sole data trough


march_data_collector/src/march_data_collector/data_collector_node.py
```diff
@@ -28,6 +32,36 @@ def __init__(self, com_calculator, cp_calculators):
...
         self._imu_subscriber = rospy.Subscriber('/march/imu', Imu, self.imu_callback)
 
+        self.pressure_soles_on = rospy.get_param('pressure_soles')
+        if self.pressure_soles_on:
+            rospy.logdebug('will run with pressure soles')
+            self.output_host = rospy.get_param('moticon_ip')
+            sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+            sock.connect(('8.8.8.8', 53))
+            self.input_host = sock.getsockname()[0]
+            sock.close()
```
Olavhaasie: Do you know why this _(Connecting to the google public DNS servers)_ is necessary if you close it immediately afterwards?

march_data_collector/src/march_data_collector/data_collector_node.py
```diff
@@ -63,5 +120,6 @@ def main():
     center_of_mass_calculator = CoMCalculator(robot, tf_buffer)
     feet = ['ankle_plate_left', 'ankle_plate_right']
     cp_calculators = [CPCalculator(tf_buffer, foot) for foot in feet]
-    DataCollectorNode(center_of_mass_calculator, cp_calculators)
+    data_collector_node = DataCollectorNode(center_of_mass_calculator, cp_calculators)
+    data_collector_node.receive_udp()
     rospy.spin()
```
Olavhaasie: You block the thread here by calling `data_collector_node.receive_udp()`. Therefore, it will never reach the `rospy.spin()` and will never be able to receive any new messages on topics. You can solve this by creating a method on the `DataCollectorNode` that does the socket loop and at the end calls `rospy.spin_once()` while ros is not shutdown. So something like this:
```python
def run(self):
    while not rospy.is_shutdown:
        self.receive_udp_once()
        rospy.spin_once()

File: march_shared_resources/package.xml [hunk 3]
```diff
@@ -18,6 +19,7 @@
   <build_export_depend>message_generation</build_export_depend>
   <build_export_depend>message_runtime</build_export_depend>
   <build_export_depend>std_msgs</build_export_depend>
+  <build_export_depend>sensor_msgs</build_export_depend>
```
Olavhaasie: I don't think you need the `sensor_msgs` package anymore, since you only use `Header`, `time` and `float32`

march_data_collector/src/march_data_collector/data_collector_node.py
RutgerVanBeek: I need the external ip-adress to bind to, because that is where the data is send to from the Moticon software. I thought it was a overkill as well, but methods like `socket.gethostbyname(socket.gethostname())` return the internal ip-address. Let me know, if you have suggestions to handle this better.

march_data_collector/src/march_data_collector/data_collector_node.py
RutgerVanBeek: I tested whether this blocked other ROS msgs. This was not the case, but I agree that your solutions s nicer.

march_data_collector/src/march_data_collector/data_collector_node.py
Olavhaasie: Okay it is how it is. You should probably also close the `input_sock` and `output_sock` when the program is done and when the `socket.error` occurs.

march_data_collector/src/march_data_collector/data_collector_node.py
```diff
@@ -48,11 +83,43 @@ def imu_callback(self, data):
...
+
+    def close_sockets(self):
+        self.output_sock.close()
+        self.input_sock.close()
+
+    def run(self):
+        while not rospy.is_shutdown():
+            if self.pressure_soles_on:
+                self.receive_udp()
+            rospy.spin_once()
```
Roelemans: This errors

PR #453, opened 2020-02-13
add version to subgait.msg

RutgerVanBeek:
Closes PM-273

## Description

There was a little bug in the subgait.py, where the version in the msg was not set, when a msg was create from the class. Therefore the version is not in the gait files and also not in the messages on /march/gait/schedule/gait 

The version is now also read from the filename and not from the file. I think it is better to also read the version from the gait files, but these are all empty strings now.
## Changes

The version is also set in the to_subgait_msg method in the subgait class.

### DIFF
--- a/march_shared_classes/src/march_shared_classes/gait/subgait.py
+++ b/march_shared_classes/src/march_shared_classes/gait/subgait.py
@@ -163,7 +163,7 @@ class Subgait(object):
         subgait_msg.trajectory = self._to_joint_trajectory_msg()
         subgait_msg.setpoints = self._to_setpoints_msg()
         subgait_msg.description = str(self.description)
-        subgait_msg.version = str(self.version)
+        subgait_msg.version = self.version
         subgait_msg.duration = rospy.Duration.from_sec(self.duration)
 
         return subgait_msg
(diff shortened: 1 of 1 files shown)


march_shared_classes/src/march_shared_classes/gait/subgait.py
```diff
@@ -163,6 +163,7 @@ def to_subgait_msg(self):
         subgait_msg.trajectory = self._to_joint_trajectory_msg()
         subgait_msg.setpoints = self._to_setpoints_msg()
         subgait_msg.description = str(self.description)
+        subgait_msg.version = str(self.version)
```
Roelemans:
```suggestion
        subgait_msg.version = self.version
```
string should not be necessary

march_shared_classes/src/march_shared_classes/gait/subgait.py
```diff
@@ -163,6 +163,7 @@ def to_subgait_msg(self):
         subgait_msg.trajectory = self._to_joint_trajectory_msg()
         subgait_msg.setpoints = self._to_setpoints_msg()
         subgait_msg.description = str(self.description)
```
Roelemans: Here too
```suggestion
        subgait_msg.description = self.description
```

PR #454, opened 2020-02-14
ESP join windows

RutgerVanBeek:
Closes PM-260

## Description

This PR joins the source windows which were created in the previous PR to one datastream. A csv output file is created. The data sources are combined, such that there is one event per joint_state event. Then for each of the other events the latest event is appended. 

You are kind of joining based on the time of the event, but because the times are not exactly equal a trick is applied, with a dummy_key with is a variable that is 1 in every event and an aggregate window which only streams the last event for every source window. 

The temperatures are first joined based on time, because the times for temperature events of different joints are always the same.  

A visualization of the model: 
image

## Changes

- Made the naming of windows more consistent. 
- Joined the temperature source windows based on time
- Joined all the sources based on a dummy key to create one source

### DIFF
--- a/march_data_collector/esp_models/march.xml
+++ b/march_data_collector/esp_models/march.xml
@@ -222,8 +222,6 @@
                 <field-selection name="orientation" source="r_orientation" />
                 <field-selection name="rotational_velocity" source="r_rotational_velocity" />
                 <field-selection name="linear_acceleration" source="r_linear_acceleration" />
-                <field-selection name="motor_current" source="l_motor_current" />
-                <field-selection name="motor_voltage" source="l_motor_voltage" />
               </output>
             </window-join>
             <window-join name="join_com" >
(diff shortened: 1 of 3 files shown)


march_data_collector/esp_models/march.xml
```diff
@@ -21,166 +22,317 @@
...
+                  <fields left="dummy_key" right="dummy_key" />
                 </conditions>
               </join>
               <output>
                 <field-selection name="time" source="l_time" />
+                <field-selection name="dummy_key" source="l_dummy_key" />
                 <field-selection name="actual_positions" source="l_actual_positions" />
                 <field-selection name="actual_velocities" source="l_actual_velocities" />
                 <field-selection name="desired_positions" source="l_desired_positions" />
                 <field-selection name="desired_velocities" source="l_desired_velocities" />
```
Olavhaasie: The window is called `join_temperature`, however the positions and velocities are also outputted?

march_data_collector/esp_models/march.xml
```diff
@@ -4,13 +4,14 @@
...
           <windows>
-            <window-join name="joinTemperature">
+            <window-join name="join_temperature_all">
               <join type="inner">
                 <conditions>
                   <fields left="time" right="time" />
                 </conditions>
               </join>
               <output>
+                <field-selection name="dummy_key" source="l_dummy_key" />
```
Olavhaasie: Why did you add all the `dummy_key`s?

march_data_collector/esp_models/march.xml
RutgerVanBeek: I mean that the temperature data is added to the base stream of the joint state data. So for every joint state event the last temperature event is appended. So the temperature is appended to the joint state event. Same for IMC, Com, IMU and so on. The last event for each of them is appended to the main event stream.

march_data_collector/esp_models/march.xml
RutgerVanBeek: You can't join the event based on time, because these are not exactly the same. However I actually want to join the events that have similar time. Therefor the aggregatewindow filters the event stream to only stream the last event. The join should then happen always, but because a key is necessary for a join I made dummy keys for every event. This method is based on a suggestion from a person who knows a lot of ESP, which I will also ask for feedback.

Olavhaasie:
Closed since #460 already contains this

PR #457, opened 2020-02-18
Add building and installing xacro as urdf

Olavhaasie:
## Description
This PR adds functionality to the building process of `march_description` it builds and installs xacro files as urdf when you build the project. I have added a CMake function that uses the `xacro_add_files` function from the `xacro` package to run `xacro` on the files and install them into install and devel space. This is useful for checking errors in the xacro file at build, so Travis will detect them. And it should no longer be necessary to process the `xacro` file during launch. So:

```xml
<param name="robot_description"
       command="$(find xacro)/xacro '$(find march_description)/urdf/march4.xacro'"/>
```
can be replaced with 
```xml
<param name="robot_description" textfile="$(find march_description)/urdf/march4.urdf"/>
```

## Changes
* Add CMake function `build_and_install_xacro_files`
* Add `xacro` dependency on `march_description`

### DIFF
--- a/march_data_collector/esp_models/march.xml
+++ b/march_data_collector/esp_models/march.xml
@@ -158,6 +158,20 @@
                 <field-selection name="com_position" source="r_com_position" />
               </output>
             </window-join>
+            <window-copy name='copyWindow'>
+              <retention field='time' type='bytime_sliding'>300 seconds</retention>
+              <connectors>
+                <connector class='fs' name='sub'>
+                  <properties>
+                    <property name='type'>sub</property>
+                    <property name='fstype'>csv</property>
+                    <property name='fsname'>result.out</property>
+                    <property name='snapshot'>true</property>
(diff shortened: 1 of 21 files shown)


Olavhaasie:
> why is the change in the launch file not included in the PR?

These changes have to be done in `project-march/hardware-interface` and `project-march/simulation` since that is where they are uploaded

RutgerVanBeek:
> > why is the change in the launch file not included in the PR?
> 
> These changes have to be done in `project-march/hardware-interface` and `project-march/simulation` since that is where they are uploaded

Also in the march_launch package there is the upload_march_urdf.launch file

Olavhaasie:
> > > why is the change in the launch file not included in the PR?
> > 
> > 
> > These changes have to be done in `project-march/hardware-interface` and `project-march/simulation` since that is where they are uploaded
> 
> Also in the march_launch package there is the upload_march_urdf.launch file

Right :slightly_smiling_face: fixed that :+1:

PR #459, opened 2020-02-18
Feature/pm 265 transition to state machine

JorisWeeda:
Closes PM-265

## Description
This PR integrates the possibility of transitioning into the state machine. It checks the given input of the IPD and checks if the requested transition is possible. If not the terminal will return a warning and the walking pattern will continue uninterrupted.

If a transition is possible the current subgait will finish and the state will exit to the walk transition state machine . This state machine  uses the transition state to execute the requested transition using the perform gait scheduler. **The transition state is not directly implemented in the healthy state machine because this would cause a consistency error in the state machine structure (due to the dynamic return state of the transition state)**

When the transition is finished correctly (succeeded return) the transition state machine will exit and the new gait will be entered. In this new gait the start state will be checked because the transition requires a possible start at a sub-gait. If this is not the case it will start at his default sub-gait (right open)

The whole structure of creating the transition gait relies on the input and output keys which can be passed between states and state machines.

## Changes
* Created transition state and state machine.
* Added necessary input and output keys within walking states and state machines
* Added callback to the idle state
* Added possibility of creating transition state in the gait perform
* Added possibility of starting at specific sub-gait.
* Check if transition is possible (using transition chain)

# Tested
* In simulation model (use test_versions-v gait directory for the right data)


march_gait_selection/src/march_gait_selection/PerformGaitAction.py
```diff
@@ -26,23 +27,33 @@ def target_gait_callback(self, subgait_goal_msg):
...
+            if trajectory_state == actionlib.GoalStatus.SUCCEEDED:
+                self.action_server.set_succeeded(trajectory_state)
+            else:
+                self.action_server.set_aborted(trajectory_state)
+
+            return True
+
+        except Exception as e:
+            rospy.logwarn('An error occurred during scheduling: {er}'.format(er=e))
+            self.action_server.set_aborted('An error occurred during scheduling: {er}'.format(er=e))
```
RutgerVanBeek: This general except now catches all kinds of different errors. For instance when I try to do a non-existing gait, I get the  catched "'NoneType' object has no attribute '__getitem__'" here instead of the more informative warning that the gait does not exist in the gait files.  Can you improve the error handling here or make the old warnings work?

JorisWeeda:
Current structure works but the transition state has a dynamic output defined when entering the transition state. This is not really desirable, SMACH uses the consistancy check to check whether the transitions match the outcomes. With movement we will define a new method and to prevent double reviewing I should first implement this.

PR #467, opened 2020-02-23
aggregate over gait

RutgerVanBeek:
Closes PM-270

## Description

This allows for creating statistics per subgait. So every time a subgait is executed the min, max and mean error and jerk are calculated. Next to that he max pressure for both feet is saved. 

To be able to do this a service is added to the healthy_sm state machine, which can be used to retrieve the current state. This was necessary, because idle states don't show up on the gait action topics. 

I have also removed almost all array types in the ESP model, because both the Streaming Analytics windows and the visualizations in jupyter lab work a lot easier with just doubles and the arrays are meant for non-fixed length variables. 

Also a logfile launch parameter is added. This way the package can be tested with a lofile, by bypassing all the simulation/exo dependencies. 
## Changes

- added an aggregate window to the ESP model
- added the logfile launch parameter
- removed almost all arrays from the ESP model. 
- added a service to get the current state from the healthy_sm state machine


march_data_collector/src/march_data_collector/esp_adapter.py
```diff
@@ -268,32 +280,36 @@ def get_time_str(timestamp):
...
-def quaternion_to_str(quaternion):
-    """Converts geometry_msgs/Quaternion to string to use in csv string for source window.
+def list_to_str(ls):
+    """Converts a list to a csv string for source window.
 
-    :param data: quaternion to convert
+    :param data: list
     """
-    return '[{x};{y};{z};{w}]'.format(x=quaternion.x, y=quaternion.y, z=quaternion.z, w=quaternion.w)
+    return ','.join([str(value) for value in ls])
```
JorisWeeda:
```suggestion
    return ','.join(str(value) for value in ls)
```

march_data_collector/src/march_data_collector/esp_adapter.py
```diff
@@ -268,32 +280,36 @@ def get_time_str(timestamp):
...
     return datetime.datetime.fromtimestamp(time).strftime('%Y-%m-%d %H:%M:%S.%f')
 
 
-def list_to_str(ls):
+def list_to_array_str(ls):
+    """Converts a list to an array formatted string for source window.
+
+    :param data: list
+    """
     return '[' + ';'.join([str(value) for value in ls]) + ']'
```
JorisWeeda:
```suggestion
    return '[' + ';'.join(str(value) for value in ls) + ']'
```

march_data_collector/src/march_data_collector/esp_adapter.py
RutgerVanBeek: The join function can only handle one iterable argument. Setting the different things to join as separate arguments is not possible.

march_data_collector/src/march_data_collector/esp_adapter.py
RutgerVanBeek: see other comment

march_data_collector/src/march_data_collector/esp_adapter.py
```diff
@@ -268,32 +280,36 @@ def get_time_str(timestamp):
...
     """
-    return '[{x};{y};{z};{w}]'.format(x=quaternion.x, y=quaternion.y, z=quaternion.z, w=quaternion.w)
+    return ','.join([str(value) for value in ls])
 
 
 def vector_to_str(vector):
     """Converts geometry_msgs/Vector3 to string to use in csv string for source window.
 
-    :param data: vector to convert
+    :param data: geometry_msgs/Vector3
```
Olavhaasie:
```suggestion
    :type vector: geometry_msgs/Vector3
```

march_data_collector/src/march_data_collector/esp_adapter.py
```diff
@@ -268,32 +280,36 @@ def get_time_str(timestamp):
...
     return '[' + ';'.join([str(value) for value in ls]) + ']'
 
 
-def quaternion_to_str(quaternion):
-    """Converts geometry_msgs/Quaternion to string to use in csv string for source window.
+def list_to_str(ls):
+    """Converts a list to a csv string for source window.
 
-    :param data: quaternion to convert
+    :param data: list
```
Olavhaasie:
```suggestion
    :type ls: list
```

march_data_collector/src/march_data_collector/esp_adapter.py
```diff
@@ -268,32 +280,36 @@ def get_time_str(timestamp):
...
-    time = timestamp.secs + timestamp.nsecs * 10**(-9)
+    time = timestamp.secs + timestamp.nsecs * 10 ** (-9)
     return datetime.datetime.fromtimestamp(time).strftime('%Y-%m-%d %H:%M:%S.%f')
 
 
-def list_to_str(ls):
+def list_to_array_str(ls):
+    """Converts a list to an array formatted string for source window.
+
+    :param data: list
```
Olavhaasie:
```suggestion
    :type data: list
```

march_data_collector/launch/march_data_collector.launch
```diff
@@ -3,9 +3,15 @@
...
+    <arg name="logfile" default="false" doc="Whether the data input is from a log file" />
     <node name="march_data_collector" pkg="march_data_collector" type="march_data_collector" output="screen">
         <param name="moticon_ip" value="$(arg moticon_ip)" />
         <param name="pressure_soles" value="$(arg pressure_soles)" />
     </node>
-    <node if="$(arg esp)" name="esp_adapter" pkg="march_data_collector" type="esp_adapter" output="screen"/>
+    <node if="$(arg esp)" name="esp_adapter" pkg="march_data_collector" type="esp_adapter" output="screen">
+        <param name="logfile" value="$(arg logfile)"/>
+    </node>
+    <group if="$(eval arg('logfile'))">
```
Olavhaasie:
```suggestion
    <group if="$(arg logfile)">
```

march_data_collector/src/march_data_collector/esp_adapter.py
```diff
@@ -154,22 +165,36 @@ def send_to_esp(self, csv, source):
         event_block = modelingApi.EventBlockNew1(event_vector, modelingApi.ebt_NORMAL)
         ret = pubsubApi.PublisherInject(pub, event_block)
         modelingApi.EventBlockDestroy(event_block)
-        return ret == 1
+        if ret != 1:
+            rospy.logwarn('Unsuccesfull inject into ESP server for source window {source} and event {event}'.format(
```
Olavhaasie:
```suggestion
            rospy.logwarn('Unsuccessful inject into ESP server for source window {source} and event {event}'.format(
```

march_data_collector/src/march_data_collector/esp_adapter.py
```diff
@@ -268,32 +280,36 @@ def get_time_str(timestamp):
...
-    :param data: vector to convert
+    :param data: geometry_msgs/Vector3
     """
-    return '[{x};{y};{z}]'.format(x=vector.x, y=vector.y, z=vector.z)
+    return '{x},{y},{z}'.format(x=vector.x, y=vector.y, z=vector.z)
 
 
 def convert_stringv(stringv, free):
-    """Converts a stringV object (integer pointer, which points to a string vector via de API) to a list.
+    """Converts a stringV object (finteger pointer, which points to a string vector via de API) to a list.
```
Olavhaasie: What is an `finteger`?

march_data_collector/src/march_data_collector/esp_adapter.py
```diff
@@ -154,22 +165,36 @@ def send_to_esp(self, csv, source):
...
+        """Callback for stopped gait. If the current state is an idle state send this to ESP a gait.
+
+        :param data: ROS message
+        :param source: the name of the source window in the ESP engine
+        """
+        rospy.sleep(0.1)
+        state = self.get_gait().current_state
+        if not ('GAIT' in state or 'HOME' in state):
+            csv = ','.join([get_time_str(data.header.stamp), 'idle', state.lower(), ' '])
+            self.send_to_esp(csv, source)
```
Olavhaasie: Perhaps it is nicer to implement this in the `state_machine` and let the service also return `is_idle_state`. Since the names of `GAIT` and `HOME` could change, it would be nicer to implement it there where someone could change it.

march_data_collector/src/march_data_collector/esp_adapter.py
```diff
@@ -154,22 +165,36 @@ def send_to_esp(self, csv, source):
...
+            rospy.logwarn('Unsuccesfull inject into ESP server for source window {source} and event {event}'.format(
+                source=source, event=csv))
+
+    def gait_finished_callback(self, data, source):
+        """Callback for stopped gait. If the current state is an idle state send this to ESP a gait.
+
+        :param data: ROS message
+        :param source: the name of the source window in the ESP engine
+        """
+        rospy.sleep(0.1)
```
Olavhaasie: Why do you sleep here?

march_data_collector/launch/march_data_collector.launch
```diff
@@ -3,9 +3,15 @@
             EMS switch laptop on standard router"/>
     <arg name="pressure_soles" default="false" doc="Whether pressure soles will be connected"/>
     <arg name="esp" default="false" doc="Whether to launch the esp adapter"/>
+    <arg name="logfile" default="false" doc="Whether the data input is from a log file" />
```
Olavhaasie: Why is the parameter named `logfile`? I cannot find a reference to reading a logfile or something like that

march_data_collector/src/march_data_collector/esp_adapter.py
```diff
@@ -67,6 +68,19 @@ def __init__(self):
...
+                """Mocks the get_gait ROS service when not available.
+
+                :return: mock CurrentState msg
+                """
+                msg = CurrentState()
+                msg.current_state = 'UNKNOWN'
+                return msg
+            self.get_gait = mock_get_gait
+        else:
+            self.get_gait = rospy.ServiceProxy('march/state_machine/current_states', CurrentState, persistent=True)
```
Olavhaasie: You could also check if the service is available and if it's not then use the mock

march_data_collector/src/march_data_collector/esp_adapter.py
RutgerVanBeek: a mistake

march_data_collector/src/march_data_collector/esp_adapter.py
RutgerVanBeek: When the gait is finished and you would directly call for the current state, it will still be that state. When you wait a bit the msg is resolved by the state machine and the new state returned. If you have a better idea on how to this, I am open for suggestions.

march_data_collector/launch/march_data_collector.launch
RutgerVanBeek: The idea is that when you set this to true that the nodes in the package do not depend on the simulation or hardware anymore. You can then play a logfile using rosbag and only launch this package. This makes testing ESP for me much easier, because I also get hardware data and don't have to run the simulation each time.

march_data_collector/src/march_data_collector/esp_adapter.py
Olavhaasie: Weird, I would think that when the gait is finished it would return that it is in a different state.

march_data_collector/src/march_data_collector/esp_adapter.py
```diff
@@ -175,10 +178,10 @@ def gait_finished_callback(self, data, source):
         :param data: ROS message
         :param source: the name of the source window in the ESP engine
         """
-        rospy.sleep(0.1)
-        state = self.get_gait().current_state
-        if not ('GAIT' in state or 'HOME' in state):
-            csv = ','.join([get_time_str(data.header.stamp), 'idle', state.lower(), ' '])
+        rospy.sleep(0.03)
+        state = self.get_gait()
+        if 'idle' in state.state_type:
```
Olavhaasie: Maybe you should check for an case insensitive `in`, because the mock returns `IDLE`, which would be false.

march_data_collector/src/march_data_collector/esp_adapter.py
```diff
@@ -67,6 +68,23 @@ def __init__(self):
...
+                """Mocks the get_gait ROS service when not available.
+
+                :return: mock CurrentState msg
+                """
+                msg = CurrentState()
+                msg.current_state = 'UNKNOWN'
+                msg.state_type = 'idle'
+                return msg
+            self.get_gait = mock_get_gait
+
```
JorisWeeda: why put this function inside the exception, I would place it outside the class or as a staticmethod to the class to prevent confusion.

march_data_collector/src/march_data_collector/esp_adapter.py
JorisWeeda: I don't think that's true, look;
image
just a minor suggestion, does not do much.

march_data_collector/src/march_data_collector/esp_adapter.py
RutgerVanBeek: The function is only used when the other one is not available, thought it would be more readable this way.

JorisWeeda: Allright.

PR #472, opened 2020-02-27
decrease max effort

RutgerVanBeek:
## Description

When looking at the short circuit errors I found that it only happens when efforts greater than 16000 are send. As far as I looked only the RKFE sends these high efforts and it is the only one with short circuit errors. Sometimes sending these efforts is okay, but a significant amount of time it results in a short circuit error. This is mainly because when you send an high effort the actual effort is often quite different (both higher and lower) sometimes resulting in doing efforts that are above the limits. The current limits are the theoretical limits, but as the effort command is not executed exactly, I think 16000 is a safe margin to make sure we do not have the short circuit error anymore. 

I don't know what the effects will be on the gaits, so may be we should airgait this first on a heavy gait. If this results in the joint not being able to follow the trajectory we might have to change the gait or come up with a real solution. 
## Changes

- lower max effort for rotational joints

### DIFF
--- a/march_description/urdf/march4.xacro
+++ b/march_description/urdf/march4.xacro
@@ -19,7 +19,8 @@
         </gazebo>
     </xacro:if>
 
-    <xacro:arg name="max_effort_rotary" default="20475" /> <!-- = 25 A -->
+    <xacro:arg name="max_effort_rotary" default="16000.0" />
+    <!-- Lowered based because of shortcircuit errors, theoretical max 20475 = 25 A -->
     <xacro:arg name="max_effort_linear" default="22932.0" />  <!-- = 28 A -->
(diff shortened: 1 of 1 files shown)


march_description/urdf/march4.xacro
```diff
@@ -19,7 +19,8 @@
         </gazebo>
     </xacro:if>
 
-    <xacro:arg name="max_effort_rotary" default="20475" /> <!-- = 25 A -->
+    <xacro:arg name="max_effort_rotary" default="16000.0" />
+    <!-- Lowered based because of shortcircuit errors, theoretical max 20475 = 25 A -->
```
Olavhaasie: Shouldn't this comment be above the `arg` :thinking:

PR #474, opened 2020-02-28
effort mode esp

RutgerVanBeek:
Closes PM-281

## Description

This PR connect the effort command and actual effort to ESP. The effort command is taken from the afterlimitjointcommand topic and the effort from the joint_states topic. I have tested the the ESP model and verified that the effort events arrived. 

I have also changed the march_launch.launch file to automatically start the march_data_collector. So when the launch parameter esp is set to true a connection with an esp server is established. 
## Changes

- Added source windows for effort and effort command to ESP model
- Added callbacks and configurations for effort and effort_command to esp_adapter
- Added the launch of the march_data_collector to the march_launch.launch

### DIFF
--- a/march_launch/launch/march.launch
+++ b/march_launch/launch/march.launch
@@ -2,6 +2,8 @@
     <arg name="configuration" default="exoskeleton" doc="Configuration on launching the 'simulation' or 'exoskeleton'."/>
     <arg name="headless" default="false" doc="Launches no GUI when true."/>
     <arg name="hardware_interface" default="true" doc="Launches the hardware interface when config is 'exoskeleton."/>
+    <arg name="gain_scheduling" default="true" doc="Whether to use gain scheduling, which changes PID values based on executed gait."/>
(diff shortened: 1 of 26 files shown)


march_launch/launch/march.launch
```diff
@@ -77,4 +82,10 @@
     <include file="$(dirname)/serial_connection.launch" if="$(arg input-device)">
         <arg name="wireless" value="$(arg wireless)"/>
     </include>
+
+    <include file="$(find march_data_collector)/launch/march_data_collector.launch">
+        <arg name="esp" value="$(arg esp)"/>
+        <arg name="pressure_soles" value="$(arg pressure_soles)"/>
+        <arg name="moticon_ip" value="$(arg moticon_ip)"/>
+    </include>
```
Olavhaasie: Would it be possible to add this to the `march` group?

RutgerVanBeek: Oopsie, thought all the groups were for different exo configurations.

PR #477, opened 2020-02-28
Set k_postion of ankles to 5.0.

Roelemans:
k position to 5 for ankles. This in necessary to disable wiggles and enable walking.


Roelemans:
But then we can't train :(

Olavhaasie:
Closing this for now, since this requires more research

Roelemans:
@Deakensbubbles I think we should merge this. 5.0 is not safe, but 0.44 is even more unsafe (wiggles). We should than find a better solution as quickly as possible.

Roelemans:
You are a hero! :tada:

PR #478, opened 2020-03-02
Load current gains from parameter sever every gain scheduling

Roelemans:
Closes PM-296

## Description

Load current gains from parameter server every time gain scheduling callback is called.

### DIFF
--- a/march_gain_scheduling/src/march_gain_scheduling/dynamic_pid_reconfigurer.py
+++ b/march_gain_scheduling/src/march_gain_scheduling/dynamic_pid_reconfigurer.py
@@ -66,8 +66,8 @@ class DynamicPIDReconfigurer:
 
     def load_current_gains(self):
         self.current_gains = []
-        for joint_index in range(len(self._joint_list)):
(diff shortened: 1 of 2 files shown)


march_gain_scheduling/src/march_gain_scheduling/dynamic_pid_reconfigurer.py
```diff
@@ -65,6 +64,12 @@ def client_update(self):
                                                                 self.current_gains[i][1],
                                                                 self.current_gains[i][2]))
 
+    def load_current_gains(self):
+        self.current_gains = []
+        for joint_index in range(len(self._joint_list)):
```
JorisWeeda: why not just use a normal for loop (also applicable for line 17!):
```suggestion
        for joint_name in self._joint_list:
```

Roelemans:
> (Should have commented this before but) how is it possible that a DynamicPIDReconfigurer can be constructed with a join_list of None? If this is the case then construction such an object would make no sense since almost all functionalities are based on this list.
Perhaps, but I think this is fine too, this has some small advantages but it doesn't really matter

PR #479, opened 2020-03-02
Add /march/gait/current topic

Olavhaasie:
## Description
In order to make my life easier for PM-292, I created a topic `/march/gait/current` in the state machine which publishes the current gait as a string. This topic does not publish when gaits have finished.

## Changes
* Add `/march/gait/current` topic to `GaitStateMachine`
* Make `StepStateMachine` inherit from `GaitStateMachine` to reduce code duplication

### DIFF
--- a/march_state_machine/src/march_state_machine/state_machines/gait_state_machine.py
+++ b/march_state_machine/src/march_state_machine/state_machines/gait_state_machine.py
@@ -1,4 +1,6 @@
+import rospy
 import smach
+from std_msgs.msg import String
 
 from march_state_machine.control_flow import control_flow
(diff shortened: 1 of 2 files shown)


Olavhaasie:
> As of now if the developer pushes a gait to be executed the publisher will always publish this gait even when the gait is not possible (for example small walk without a valid small walk in the gait directory). Are u aware of this?

I was not aware of this. Right now it publishes it on the topic right when it enters the state machine and that is before any of the subgaits are executed

> And don't u want to publish if the gait has been finished on the same topic?

I can probably add this by publishing the message once the state machine for a gait exits

PR #481, opened 2020-03-03
Fix: interpolation steps

Olavhaasie:
Closes PM-304

## Description
This PR fixes the issue where the interpolation would take a longer amount of time than was expected. It turns out that the `last_update_time` was used for every joint and updated on every joint, which resulted in the time_interval being very small. Besides this I also made some changes to improve the performance of the interpolation loop, by moving some parameter queries outside of the loop. The loop can now run at around 6Hz (The Dynamic Reconfigure client update is very slow) and the interpolation for `walk_like` and `sit_like` is now done in ~1.5 seconds.

## Changes
* Add a last update time for every joint
* Move parameter queries outside interpolate loop
* Change slope to 100

### DIFF
--- a/march_gain_scheduling/src/march_gain_scheduling/dynamic_pid_reconfigurer.py
+++ b/march_gain_scheduling/src/march_gain_scheduling/dynamic_pid_reconfigurer.py
@@ -12,7 +12,8 @@ class DynamicPIDReconfigurer:
         self._joint_list = joint_list
         self.interpolation_done = True
         self._last_update_times = []
-        self._clients = [Client('/march/controller/trajectory/gains/' + j, timeout=30) for j in self._joint_list]
+        self._clients = [Client('/march/controller/trajectory/gains/' + joint, timeout=30) for joint in
+                         self._joint_list]
         self.current_gains = []
         rospy.Subscriber('/march/gait/schedule/goal', GaitActionGoal, callback=self.gait_selection_callback)
         self._linearize = rospy.get_param('~linearize_gain_scheduling')
--- a/march_gain_scheduling/src/march_gain_scheduling/dynamic_pid_reconfigurer.py
+++ b/march_gain_scheduling/src/march_gain_scheduling/dynamic_pid_reconfigurer.py
@@ -7,63 +7,60 @@ from .one_step_linear_interpolation import interpolate
 
 
 class DynamicPIDReconfigurer:
-    def __init__(self, joint_list=None, max_time_step=0.1):
+    def __init__(self, joint_list):
         self._gait_type = None
         self._joint_list = joint_list
-        self._max_time_step = max_time_step
         self.interpolation_done = True
-        self.last_update_time = None
-        self._clients = []
+        self._last_update_times = []
+        self._clients = [Client('/march/controller/trajectory/gains/' + j, timeout=30) for j in self._joint_list]
         self.current_gains = []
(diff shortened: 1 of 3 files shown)


Deakensbubbles:
In the spirit of our sprint planning on monday; has this been tested on the exo yet?

Olavhaasie:
> In the spirit of our sprint planning on monday; has this been tested on the exo yet?

I have added a new label with which you can indicate if you want this tested

march_gain_scheduling/src/march_gain_scheduling/dynamic_pid_reconfigurer.py
```diff
@@ -7,63 +7,60 @@
...
-    def __init__(self, joint_list=None, max_time_step=0.1):
+    def __init__(self, joint_list):
         self._gait_type = None
         self._joint_list = joint_list
-        self._max_time_step = max_time_step
         self.interpolation_done = True
-        self.last_update_time = None
-        self._clients = []
+        self._last_update_times = []
+        self._clients = [Client('/march/controller/trajectory/gains/' + j, timeout=30) for j in self._joint_list]
```
Roelemans:
```suggestion
        self._clients = [Client('/march/controller/trajectory/gains/' + joint, timeout=30) for joint in self._joint_list]
```

march_gain_scheduling/src/march_gain_scheduling/dynamic_pid_reconfigurer.py
```diff
@@ -7,63 +7,60 @@
...
 
         if self._gait_type != new_gait_type or not self.interpolation_done:
-            rospy.logdebug('The selected gait: {0} is not the same as the previous gait: {1}'.format(
-                new_gait_type, self._gait_type))
-
-            self.load_current_gains()
             self._gait_type = new_gait_type
+            self.load_current_gains()
             self.interpolation_done = False
+            self._last_update_times = len(self._joint_list) * [rospy.get_time()]
```
Roelemans: move to behind line 37 and input begin_time directly? If the get_param in lookup_table takes very long, this might not result in ideal behavior.
```suggestion
            self._last_update_times = len(self._joint_list) * [begin_time]
```

march_gain_scheduling/src/march_gain_scheduling/dynamic_pid_reconfigurer.py
Olavhaasie: I did this because the line would otherwise be too long

Roelemans: @Olavhaasie fixed

Olavhaasie:
> I get a Key error for when the gain scheduling node tries to get the joint names from the parameter server. I did a beunfix on exo with a sleep.

Fixed this by waiting for the parameter be available.

> Furthermore when I test this in simulation the scheduling does not start during homing, but at the next gait.

I cannot reproduce this. Make sure you launch using:

    roslaunch march_launch march_simulation.launch gain_scheduling:=true

How do you know it didn't interpolate on the first gait?

RutgerVanBeek:
> How do you know it didn't interpolate on the first gait?

I listened to /march/controller/trajectory/gains/left_ankle/parameter_updates. I can also not reproduce it anymore so it seems alright.

PR #482, opened 2020-03-03
add argument to urdf from parameter server

RutgerVanBeek:
## Description

Don't know what changed in uploading of the URDF, but this fixes the data_collector not being able to get the URDF anymore. 
## Changes

* Added an argument to get_from_parameter_server

### DIFF
--- a/march_data_collector/src/march_data_collector/data_collector_node.py
+++ b/march_data_collector/src/march_data_collector/data_collector_node.py
@@ -151,7 +151,7 @@ class DataCollectorNode(object):
 def main():
     rospy.init_node('data_collector', anonymous=True)
     try:
-        robot = URDF.from_parameter_server()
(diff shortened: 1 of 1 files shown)


Olavhaasie:
Ah I think I know why this happened. It defaults to getting it from the `robot_description` parameter in its own namespace. However, in commit https://github.com/project-march/march/pull/474/commits/c880ce9fc2c47eb7a6a591e38840286bcd80e5d5 you added it to the `march` namespace, which caused it to look for `/march/robot_description`, which does not exist. _(I told you that lol)_ :sweat_smile:

Ishadijcks:
Is there a reason the `robot_description` is uploaded at `/robot_description` and not at `/march/robot_description`?

Olavhaasie:
> Is there a reason the `robot_description` is uploaded at `/robot_description` and not at `/march/robot_description`?

Not necessarily

PR #488, opened 2020-03-09
Feature/pm 305 add walk large and small to transition

JorisWeeda:
Closes PM-305

## Description
Currently the transition state is only available when the walk medium is selected. The walk large and walk small start states could not use the transition state because these were separate states in the healthy state machine. This PR creates the possibility to use the walk_small and walk_large as start states and transition to other states when requested.

## Changes
* Added custom start state to the transition state machine
* Deleted separate walk_small and walk_large from the healthy state machine
* Created start state of walk_small and walk_large which are linked to the transition state machine


march_state_machine/src/march_state_machine/healthy_sm.py
```diff
@@ -47,14 +48,13 @@ def __init__(self):
...
-        walking_state_machine.add('walk', WalkStateMachine('walk', is_transition_active=True), default_start=True)
         walking_state_machine.add('walk_large', WalkStateMachine('walk_large', is_transition_active=True))
 
-        self.add_state('GAIT WALK', walking_state_machine, 'STANDING')
-
-        self.add_state('GAIT WALK SMALL', WalkStateMachine('walk_small'), 'STANDING')
-        self.add_state('GAIT WALK LARGE', WalkStateMachine('walk_large'), 'STANDING')
+        self.add_state('GAIT WALK', walking_state_machine, 'STANDING', custom_start_label='walk')
+        self.add_state('GAIT WALK SMALL', walking_state_machine, 'STANDING', custom_start_label='walk_small')
+        self.add_state('GAIT WALK LARGE', walking_state_machine, 'STANDING', custom_start_label='walk_large')
```
Olavhaasie: In your PR description you mentioned that you removed the small and large gaits from the healthy state machine, however, here you eve add the normal walking gait. Why are these still necessary? Wouldn't it be a lot nicer to implement this with input/output keys from smach? Where the output key of the `IDLE STATE` could be the size of the walk?

march_state_machine/src/march_state_machine/healthy_sm.py
```diff
@@ -189,3 +194,10 @@ def get_current_states(self, _req):
         state = self.get_active_states()[0]
         state_type = str(type(self[state]))
         return state_type, state
+
+    def _update_once(self):
+        if self._current_label in self._custom_start_states:
+            if isinstance(self._current_state, StateMachineWithTransition):
+                self._current_state.initial_state_label = self._custom_start_states[self._current_label]
```
Olavhaasie: I am not 100% sure, but I think you can do this a lot nicer using a transition callback from the state machine. Take a look at `register_transition_cb` in the `smach.Container`. this callback will be called on every transition in the healthy statemachine (I believe not in sub state machines).

march_state_machine/src/march_state_machine/state_machines/transition_state_machine.py
```diff
@@ -52,12 +50,42 @@ def execute(self, parent_ud=smach.UserData()):
...
+        if self.initial_state_label not in self._transition_sequence:
+            return 'succeeded'
+
+        return super(StateMachineWithTransition, self).execute(parent_ud)
+
+    def _update_once(self):
+        if self.initial_state_label not in self._transition_sequence:
+            return 'succeeded'
+
+        return super(StateMachineWithTransition, self)._update_once()
```
Olavhaasie: Why is this override necessary? This is also checked in the `execute` method and that is the only place where this can occur right?

march_state_machine/src/march_state_machine/healthy_sm.py
JorisWeeda: What I implied in my description is that I removed the old walk small and walk large buttons because these were stand alone state machines which were not linked to the transition. The new walk small an walk large states are actually the transition state machine but with a custom start. The normal walk was already in the file but just changed line.

march_state_machine/src/march_state_machine/healthy_sm.py
JorisWeeda: This indeed looks like a better solution!

march_state_machine/src/march_state_machine/state_machines/transition_state_machine.py
JorisWeeda: I think I forgot to remove this override method, it is indeed redundant.

march_state_machine/src/march_state_machine/healthy_sm.py
Olavhaasie: So now there exist 3 different transition state machines with each a different start state?

JorisWeeda: No, there is a single transition state machine (which has the walk_small, walk_medium, walk_large state machines) and the GAIT WALK, GAIT WALK SMALL and GAIT WALK LARGE all refer to this transition state but each with a custom start state.

Olavhaasie: Will you change the implementation or should we leave it? It does work

JorisWeeda: I'm not sure, it does work pretty nice but it seems redundant to have three identical states. I've shortly looked into other options but cant seems to find anything for now. My initial thought is to keep it like this and I will create a JIRA issue to remove the identical states, u agree?

Olavhaasie: Agreed :+1:

PR #490, opened 2020-03-11
Fix safety node ipd reconnection messages

Olavhaasie:
Closes PM-317
## Description
Fixed the message on the safety node that was supposed to publish when a input device was reconnected.

It will now publish the non fatal once, when it loses a connection, which will try to stop the current gait. Furthermore, it will print once an input device was reconnected.

## Changes
* Publish non fatal input device connection lost once
* Add log info when an input device reconnects
* Change ping from future to warning
* Lower cycle rate on safety node from 200 to 20


march_safety/src/input_device_safety.cpp
```diff
@@ -21,45 +21,53 @@ InputDeviceSafety::InputDeviceSafety(ros::NodeHandle* n, SafetyHandler* safety_h
...
-  if (ros::Time::now() > this->time_last_send_error_ + ros::Duration(this->send_errors_interval_ / 1000.0))
+
+  // Check if there is no alive msg receive for the timeout duration.
+  if (now > this->time_last_alive_ + this->connection_timeout_)
+  {
+    this->is_connected_ = false;
+    std::ostringstream message_stream;
+    message_stream << "Input Device Connection Lost. Current time is " << now.toSec() << " and last alive message was "
+                   << this->time_last_alive_.toSec()
+                   << ". The difference in time is: " << now.toSec() - this->time_last_alive_.toSec();
```
Roelemans: Is it possible to make this the kind of message that updates? like one message with a running timer (like for example in colcon build)

march_safety/src/input_device_safety.cpp
```diff
@@ -21,45 +21,53 @@ InputDeviceSafety::InputDeviceSafety(ros::NodeHandle* n, SafetyHandler* safety_h
...
+    std::string error_message = message_stream.str();
```
Roelemans: You get this message twice: once directly and once because "SafetyState has noticed a non fatal error: Input Device Connection Lost. Current ..." and followed by a warning "Idle state does not respond to stop". Maybe we should switch off the error message in the SafetyState?

march_safety/src/input_device_safety.cpp
```diff
@@ -21,45 +21,53 @@ InputDeviceSafety::InputDeviceSafety(ros::NodeHandle* n, SafetyHandler* safety_h
...
+
+    this->sendError(now, error_message);
+  }
+  else if (!this->is_connected_)
+  {
+    this->is_connected_ = true;
+    ROS_INFO("Input device reconnected");
```
Roelemans: This is also send when the input device connects the first time, but I personally think that is fine.

march_safety/src/input_device_safety.cpp
Olavhaasie: I believe this is already an issue on Jira: PM-272

march_safety/src/input_device_safety.cpp
Olavhaasie: I have changed it. It now publishes the non fatal once, when it loses the input device, since there is no reason to keep publishing it and trying to stop the current gait. Then it will throttle print that no input device is connected and when it reconnects it logs again that it has found a connection. The error message is still double logged.

Olavhaasie:
> The alarm that the input device is disconnected is no longer played continuously. Is that a problem?

Created issue PM-332 to add sound on input device reconnect

PR #491, opened 2020-03-12
Feature/straight tilted path mirrored

gaiavdh:
## Description
 We made a Tilted Path gait for when the obstacle has another orientation during the Cybathlon. As the other leg will be bent in this gait, we added new idle states and therefore a new state machine for this version of the obstacle.

## Changes
 New gaits in state machine with new idle states.


gaiavdh:
> Does this mean that the Tilted Path sideways can be removed?

No, it doesn't. We still have not been able to test if Sjaan can walk straight over the Tilted Path, so we still need the sideways version as a back-up. (We have not made a version of the sidestep with the different orientation yet as we first want to see if the sidestep is even needed.)

PR #501, opened 2020-03-26
Fix: unallowed gaits

Olavhaasie:
Closes PM-307

## Description
This fixes the issue where gaits executed in the state machine that are not loaded in gait selection will transition the state machine to `UNKNOWN`. With this fix the gait state machine will return a `rejected` outcome and the state machine will recover from it and go back to the original state. I have implemented this as a service on the gait selection that enables the state machine to ask whether a gait with subgaits is currently loaded. At first I thought it would be possible to make the action server in gait selection reject the action request, however this was not possible with the `SimpleActionServer` and required implementing your own `ActionServer` from actionlib. See https://github.com/ros/actionlib/issues/38 for more info.

## Changes
* Implement `/march/gait_selection/contains_gait` service
* Add `ContainsGait.srv` description
* Add `rejected` outcome to gait state machines and transition state machine

### DIFF
--- a/march_gait_selection/src/march_gait_selection/gait_selection_node.py
+++ b/march_gait_selection/src/march_gait_selection/gait_selection_node.py
@@ -4,7 +4,7 @@ import rospy
 from std_srvs.srv import Trigger
 
 from march_shared_classes.exceptions.gait_exceptions import GaitError
-from march_shared_resources.srv import ContainsGait, ContainsGaitResponse, ContainsGaitRequest, StringTrigger
+from march_shared_resources.srv import ContainsGait, ContainsGaitResponse, StringTrigger
 
 from .gait_selection import GaitSelection
 from .perform_gait_action import PerformGaitAction
--- a/march_gait_selection/src/march_gait_selection/gait_selection_node.py
+++ b/march_gait_selection/src/march_gait_selection/gait_selection_node.py
@@ -4,11 +4,10 @@ import rospy
 from std_srvs.srv import Trigger
 
 from march_shared_classes.exceptions.gait_exceptions import GaitError
-from march_shared_resources.srv import StringTrigger
+from march_shared_resources.srv import ContainsGait, ContainsGaitResponse, ContainsGaitRequest, StringTrigger
 
 from .gait_selection import GaitSelection
 from .perform_gait_action import PerformGaitAction
-
 
 NODE_NAME = 'gait_selection'
 GAIT_FILES_MAP_NAME = 'march_gait_files'
@@ -32,6 +31,26 @@ def set_gait_version_map(msg, gait_selection):
         return [False, 'Error occurred when constructing gaits: {er}'.format(er=e)]
 
 
+def contains_gait(request, gait_selection):
+    """
+    Checks whether a gait and subgait are loaded.
+
+    :type request: ContainsGaitRequest
+    :param request: service request
+    :param gait_selection: current loaded gaits
+    :return: True when the gait and subgait are loaded
+    """
+    gait = gait_selection[request.gait]
+    if gait is not None:
+        for subgait in request.subgaits:
+            if gait[subgait] is None:
+                return ContainsGaitResponse(False)
+    else:
+        return ContainsGaitResponse(False)
+
+    return ContainsGaitResponse(True)
+
+
 def main():
     rospy.init_node(NODE_NAME)
     gait_package = rospy.get_param('~gait_package', GAIT_FILES_MAP_NAME)
@@ -51,6 +70,9 @@ def main():
 
     rospy.Service('/march/gait_selection/update_default_versions', Trigger,
                   lambda msg: gait_selection.update_default_versions())
+
+    rospy.Service('/march/gait_selection/contains_gait', ContainsGait,
+                  lambda msg: contains_gait(msg, gait_selection))
 
     PerformGaitAction(gait_selection)
     rospy.spin()
--- a/march_gait_selection/src/march_gait_selection/gait_selection_node.py
+++ b/march_gait_selection/src/march_gait_selection/gait_selection_node.py
@@ -41,12 +41,11 @@ def contains_gait(request, gait_selection):
     :return: True when the gait and subgait are loaded
     """
     gait = gait_selection[request.gait]
-    if gait is not None:
-        for subgait in request.subgaits:
-            if gait[subgait] is None:
-                return ContainsGaitResponse(False)
-    else:
+    if gait is None:
         return ContainsGaitResponse(False)
+    for subgait in request.subgaits:
+        if gait[subgait] is None:
+            return ContainsGaitResponse(False)
 
     return ContainsGaitResponse(True)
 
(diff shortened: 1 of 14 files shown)


march_gait_selection/src/march_gait_selection/gait_selection_node.py
```diff
@@ -32,6 +31,26 @@ def set_gait_version_map(msg, gait_selection):
...
+    :param gait_selection: current loaded gaits
+    :return: True when the gait and subgait are loaded
+    """
+    gait = gait_selection[request.gait]
+    if gait is not None:
+        for subgait in request.subgaits:
+            if gait[subgait] is None:
+                return ContainsGaitResponse(False)
+    else:
+        return ContainsGaitResponse(False)
```
Roelemans:
```suggestion
    if gait is None:
        return ContainsGaitResponse(False)
    for subgait in request.subgaits:
        if gait[subgait] is None:
            return ContainsGaitResponse(False)
```

march_state_machine/src/march_state_machine/gaits/tilted_path_left_flexed_knee_step_sm.py
```diff
@@ -5,16 +5,17 @@
...
 
 def create():
-    sm_tilted_path_left_flexed_knee_step = smach.StateMachine(outcomes=['succeeded', 'preempted', 'failed'])
-    sm_tilted_path_left_flexed_knee_step .register_io_keys(['sounds'])
+    sm_tilted_path_left_flexed_knee_step = smach.StateMachine(outcomes=['succeeded', 'preempted', 'failed', 'rejected'])
+    sm_tilted_path_left_flexed_knee_step.register_io_keys(['sounds'])
     with sm_tilted_path_left_flexed_knee_step:
         smach.StateMachine.add('GAIT TP LEFT FLEXED KNEE STEP', StepStateMachine('tilted_path_left_flexed_knee_step',
-                               subgaits=['right_open', 'left_close']),
-                               transitions={'succeeded': 'STANDING TP LEFT STRAIGHT', 'failed': 'failed'})
```
Roelemans: Why is the failed no longer necessary?

march_state_machine/src/march_state_machine/gaits/tilted_path_right_flexed_knee_step_sm.py
```diff
@@ -5,16 +5,19 @@
...
-    sm_tilted_path_right_flexed_knee_step = smach.StateMachine(outcomes=['succeeded', 'preempted', 'failed'])
+    sm_tilted_path_right_flexed_knee_step = smach.StateMachine(
+        outcomes=['succeeded', 'preempted', 'failed', 'rejected'])
     sm_tilted_path_right_flexed_knee_step.register_io_keys(['sounds'])
     with sm_tilted_path_right_flexed_knee_step:
         smach.StateMachine.add('GAIT TP RIGHT FLEXED KNEE STEP', StepStateMachine('tilted_path_right_flexed_knee_step',
-                               subgaits=['left_open', 'right_close']),
-                               transitions={'succeeded': 'STANDING TP RIGHT STRAIGHT', 'failed': 'failed'})
+                                                                                  subgaits=['left_open',
+                                                                                            'right_close']),
```
Roelemans: Wouldn't it be neater to just put the StepStateMachine on the next line instead of splitting the subgaits? Doesn't really matter though :smiley_cat:

march_state_machine/src/march_state_machine/healthy_sm.py
```diff
@@ -167,22 +170,26 @@ def __init__(self):
...
         :param succeeded: name of the state that the given state should transition to once succeeded
+        :type rejected: str
+        :param rejected: name of the state that the given state should transition to once rejected
         :type custom_start_label: str
         :param custom_start_label: name of the start state within the given state machine
         """
         self.assert_opened()
-        self.add(label, state, transitions={'succeeded': succeeded, 'failed': 'UNKNOWN'})
+        if rejected is None:
+            rejected = succeeded
```
Roelemans: Why would you set the outcome of rejected to succeeded by default?

march_state_machine/src/march_state_machine/state_machines/gait_state_machine.py
```diff
@@ -53,6 +59,16 @@ def add_subgait(self, subgait_name, succeeded='succeeded', stopped=None):
                                    transitions={'succeeded': succeeded, 'aborted': 'failed'})
 
     def execute(self, ud=smach.UserData()):
+        try:
+            subgaits = self.get_children().keys()
+            if not self._contains_gait(gait=self._gait_name, subgaits=subgaits).contains:
+                rospy.logwarn('Gait {0} is not currently loaded with subgaits {1}'.format(self._gait_name, subgaits))
+                return 'rejected'
```
Roelemans: When in develop and training-v gait folder; when I click walk_large I get 
"[WARN] [1585237394.453210, 7.478000]: State walk_large does not have a valid gait file, removing state from transition state machine
but not the warning that you print here, whereas when I click  gait_ramp_door_last_step, I do get this error.

march_state_machine/src/march_state_machine/healthy_sm.py
JorisWeeda: I think because succeeded returns to the idle state after a state/state machine has finished

march_state_machine/src/march_state_machine/state_machines/gait_state_machine.py
JorisWeeda: The walking pattern uses the transition state machine which has different error handling. However, also the method is different so maybe (in another PR) this should be updated to using one consistent method?

march_state_machine/src/march_state_machine/gaits/tilted_path_left_flexed_knee_step_sm.py
Olavhaasie: Since the mapping of `failed` should go to the output of `failed` on the state machine, SMACH is able to automatically do this. The same goes for `preempted` and `succeeded` on the last state.

march_state_machine/src/march_state_machine/healthy_sm.py
Olavhaasie: Most gaits go from `STANDING` to `STANDING`, so the `succeeded` is the same state as the one they came from. So when a gait gets rejected it should return to its origin state. Therefore, defaulting `rejected` to `succeeded` makes sense for most states.

march_state_machine/src/march_state_machine/state_machines/gait_state_machine.py
Olavhaasie: I noticed this too, but since the transition state machine requires some changes to be more optimal I decided to leave it for now.

JorisWeeda: I agree, i shall make another jira issue for this specific problem

Roelemans: :+1:

march_state_machine/src/march_state_machine/gaits/tilted_path_right_flexed_knee_step_sm.py
Roelemans: And this?

Olavhaasie: Sorry forgot to comment on this. CLion refactored it like this and this seems more consistent with the other formatting in the files

PR #504, opened 2020-04-03
Feature/pm 393 correct height imu tf

RutgerVanBeek:
Closes PM-393

## Description

This makes sure that when the robot to world transform is made by the IMU that the correct height is set. The height is chosen such that the lowest foot is on the ground. This worked better then I expected, when I moved the IMU quite a lot, it felled really natural. The body still under the ground when you flip the IMU over but I think it does not makes sense the cover that situation, because if exo is upside down we have a lot bigger problems. 
## Changes

* Changed the transformation calculated based on the IMU to take the height of the feet into account and place the exo on the ground. 

## Testing
In the bag find there is some IMU data to test this.

### DIFF
--- a/march_data_collector/src/march_data_collector/data_collector_node.py
+++ b/march_data_collector/src/march_data_collector/data_collector_node.py
@@ -22,15 +22,12 @@ from .cp_calculator import CPCalculator
 
 
 class DataCollectorNode(object):
-    def __init__(self,robot, com_calculator, cp_calculators, tf_buffer, feet):
-        self.robot = robot
+    def __init__(self, com_calculator, cp_calculators, tf_buffer, feet):
         self.differentiation_order = 2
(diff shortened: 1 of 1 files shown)


Olavhaasie:
Is there a ROS way of simulating an IMU?

RutgerVanBeek:
> Is there a ROS way of simulating an IMU?

You mean in Gazebo or for testing this?
If you really want to test it, I can record some IMU data and store it in an bag file.

Olavhaasie:
> If you really want to test it, I can record some IMU data and store it in an bag file.

Could you record some data and share it here? :innocent: I would like to see it

RutgerVanBeek:
[2020-04-07-16-34-57.zip](https://github.com/project-march/march/files/4445205/2020-04-07-16-34-57.zip)

PR #505, opened 2020-04-06
Solve the 'random' closing of the exoskeleton

JorisWeeda:
Closes PM-390

## Description
The reset variable was not reset if the left close or right close was being executed. This way the exoskeleton could 'randomly' stop in the next walking sequence. Resetting the stop every time the idle state is entered should solve this problem

## Changes
* Reset stop variable when idle state is entered


Ishadijcks:
@TimBuckers take it as a compliment :yum:

PR #506, opened 2020-04-07
Feature/pm 391 set state machine to unknown

JorisWeeda:
Closes PM-391
(Also use the feature/PM-391-sm-to-unknown-button in the march_rqt_input_device)

## Description
This PR solves the problem to shutdown the exoskeleton in order to home-stand of home-sit again. It does this by manually setting the state to the 'UNKNOWN' state in the exoskeleton. It is only useful to have this callback in the idle_state because this state is used in the side and ramp up gaits. In order to move to the 'UNKNOWN' state the 'failed' outcome is used for every idle state and has been added if it was not available yet

## Changes
* At UNKNOWN instruction to the GaitInstruction.msg
* Added a callback method in the control_flow.py
* Added the 'failed' and transition to UNKNOWN tag
* Used the callback in the idle_state

### DIFF
--- a/march_state_machine/src/march_state_machine/gaits/slope_down_sm.py
+++ b/march_state_machine/src/march_state_machine/gaits/slope_down_sm.py
@@ -12,7 +12,8 @@ def create():
         smach.StateMachine.add('GAIT RD SLOPE DOWN', SlopeStateMachine('ramp_door_slope_down'),
                                transitions={'succeeded': 'STANDING SLOPE DOWN'})
 
-        smach.StateMachine.add('STANDING SLOPE DOWN', IdleState(outcomes=['gait_ramp_door_last_step', 'preempted']),
+        smach.StateMachine.add('STANDING SLOPE DOWN', IdleState(outcomes=['gait_ramp_door_last_step', 'preempted',
+                                                                          'failed']),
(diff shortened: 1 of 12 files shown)


march_state_machine/src/march_state_machine/gaits/slope_down_sm.py
```diff
@@ -12,7 +12,8 @@ def create():
         smach.StateMachine.add('GAIT RD SLOPE DOWN', SlopeStateMachine('ramp_door_slope_down'),
                                transitions={'succeeded': 'STANDING SLOPE DOWN'})
 
-        smach.StateMachine.add('STANDING SLOPE DOWN', IdleState(outcomes=['gait_ramp_door_last_step', 'preempted']),
+        smach.StateMachine.add('STANDING SLOPE DOWN', IdleState(outcomes=['gait_ramp_door_last_step', 'preempted',
+                                                                          'failed']),
```
Olavhaasie, line 16: I think it would be better to include the `preempted` and `failed` outcomes to the `IdleState` by default, since they all need it, there is no use in giving them as parameters. Of course, still keep the gaits, since those must be chosen and make the `outcomes` more like a `gaits` parameter.

PR #509, opened 2020-04-14
Feature controller statistics in esp

RutgerVanBeek:
Closes PM-284

## Description

I have acces to ESP studio now, this gives some more overview on the possibilities and some more debug information. I can give you a tour sometime if you would like that. This resulted in a refactor. The model is now split into two continuous queries (`temperature_join`, `gait_analysis`, `control_analysis`). Because of this the ESP adapter is adjusted to be able to inject the same event in different source windows (probably in different continuous queries). The controller analysis query will try the has events that almost all are injected at 250 Hz and therefore tries to join them at this time instead of using the dummy key. 

With this PR also the controller statistics are added to the output of the model. These are both outputted per second and per subgait (or idle state). The statistics include:
* Min and max error
* Mean squared error
* Mean squared effort change (measure for the effort peaks)
* Average amount of time the safety limits worked
* Average amount the error changed sign (measure for the wiggles)
* Correlation between the effort command and actual effort done by the iMC.

## Changes

* Refactored ESP model
* Added support for multiple source windows per rostopic
* Added controller statistics to the ESP model.


RutgerVanBeek:
> Looks good, as far as I can tell sweat_smile. Maybe it is nice to also display the average amount of time the safety limits increased effort or changed its sign?

Yes, I will implement that this afternoon or tomorrow.

Issue #256, opened 2019-07-01
rqt_input_device publishes wrong error msg


Ishadijcks:
This error is caused by the input device not publishing the same message as the safety node expects. So when you press the error button it freezes.

```
Traceback (most recent call last):
  File "/home/march/Documents/march_ws/src/march-iv/march_rqt_input_device/src/march_rqt_input_device/input_device.py", line 88, in <lambda>
    callback=lambda: self.publish_error())
  File "/home/march/Documents/march_ws/src/march-iv/march_rqt_input_device/src/march_rqt_input_device/input_device.py", line 152, in publish_error
    self.error_pub.publish(Error(-1, "Fake error thrown by the develop input device.", Error.FATAL))
  File "/home/march/Documents/march_ws/devel/lib/python2.7/dist-packages/march_shared_resources/msg/_Error.py", line 43, in __init__
    super(Error, self).__init__(*args, **kwds)
  File "/opt/ros/kinetic/lib/python2.7/dist-packages/genpy/message.py", line 294, in __init__
    raise TypeError("Invalid number of arguments, args should be %s"%str(self.__slots__)+" args are"+str(args))
TypeError: Invalid number of arguments, args should be ['error_message', 'type'] args are(-1, 'Fake error thrown by the develop input device.', 0)
```

Hints:

- Check the `march_rqt_input_device` package on how to launch it.
- Check what the safety node expects.
- Fix the input device

PR #510, opened 2020-04-15
Add motor voltage data to esp

Ciliaclaij:
Dependent on Feature/pm 300 add motor voltage #256

## Description

Added motor voltage reading to esp

## Changes

* Added motor voltage reading to esp


RutgerVanBeek:
You could not know this, but the ESP model should be slightly adjusted for this to work. But in another PR the model is totally refactored. I therefore suggest to to only adjust the message here. I will fix it in the other PR.

Ciliaclaij:
I kind of overlooked the comments on this PR, sorry for that. So I can revert the changes in the  march_data_collector/src/march_data_collector/esp_adapter.py file?

RutgerVanBeek:
> I kind of overlooked the comments on this PR, sorry for that. So I can revert the changes in the march_data_collector/src/march_data_collector/esp_adapter.py file?

Yes, exactly!

PR #512, opened 2020-04-17
Feature/blc 27 moveit integration

JorisWeeda:
Closes BLC-27

## Description
Created a first version of a MoveIt integration class. This class will eventually communicate with the capture points generated in the data collector node and use these to calculate a trajectory which can be executed by the gait scheduler. The current class only has a method to create a random sub-gait message using the motion planning functionalities of MoveIt.

## Changes
* new file in the gait selection to communicate with MoveIt
* Renamed the folder in the gait selection to 'dynamic gaits'
* Created the initialization to communicate with MoveIt and the capture points
* Created a function to generate a random subgait message using moveit motion planning


march_description/urdf/march4.xacro
```diff
@@ -758,4 +758,7 @@
             <mechanicalReduction>200</mechanicalReduction>
         </actuator>
     </transmission>
+
+
+
```
RutgerVanBeek, line 763: This should be fixed

march_gait_selection/src/march_gait_selection/dynamic_gaits/moveit_subgait.py
```diff
@@ -0,0 +1,66 @@
...
+
+        :return:
+            A populated subgait message
+        """
+        self._move_group['left_leg'].set_random_target()
+        trajectory_plan = self._move_group['left_leg'].plan()
+        return self.to_subgait_msg_('Random moveit subgait', trajectory_plan.joint_trajectory)
+
+    @staticmethod
+    def to_subgait_msg_(name, trajectory, gait_type='walk_like', version='moveit',
```
Roelemans, line 54:
```suggestion
    def to_subgait_msg_(name, trajectory_msg, gait_type='walk_like', version='moveit',
```

march_gait_selection/src/march_gait_selection/dynamic_gaits/moveit_subgait.py
```diff
@@ -0,0 +1,66 @@
...
+                        description='Subgait created using the moveit motion planning.'):
+        """Create a subgait message using the standard format in the march shared resources."""
+        subgait_msg = Subgait()
+
+        subgait_msg.name = name
+        subgait_msg.description = description
+        subgait_msg.trajectory = trajectory
```
Roelemans, line 61:
```suggestion
        subgait_msg.trajectory = trajectory_msg
```

march_gait_selection/src/march_gait_selection/dynamic_gaits/moveit_subgait.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def random_subgait(self):
+        """Create random trajectory using the moveit motion planner.
+
+        :return:
+            A populated subgait message
+        """
+        self._move_group['left_leg'].set_random_target()
+        trajectory_plan = self._move_group['left_leg'].plan()
+        return self.to_subgait_msg_('Random moveit subgait', trajectory_plan.joint_trajectory)
```
Roelemans, line 51:
```suggestion
        trajectory_msg = self._move_group['left_leg'].plan()
        return self.to_subgait_msg_('Random moveit subgait', trajectory_msg.joint_trajectory)
```

JorisWeeda:
> I have a few questions:
> 
>     * Why are there so many launch files? Do they all have a purpose? Is it correct that only `setup_assistant.launch` is installed?
From what I know is the setup assistant creates all the launch files and each has his own individual function. Looking at the ```move_group.launch``` some are necessary to access the motion planner we use to define the trajectory. Some are redundant and used for the demo launch files.
>     * Why do some files have _fake_ in the name? Should they be pushed?
The 'fake' named launch files are mostly used for visualization and testing with gazebo and rviz 
>     * Don't you think it is better to include the SRDF file in the `march_description` package? Since it is a description for the march robot and it should probably be modified together with the URDF.
We could do this, my preference would be to first get the MoveIt class up and running and after that clean and remove unnecessary files and maybe move some data also. This may be backwards to what we are used to but in my opinion were are not yet a 100% sure of which MoveIt packages and launch files. Would that be something u would agree on? 
>     * Is there some command I can run to see this work?
I've slacked u the details!

Maybe @RutgerVanBeek (and @Roelemans) would also share their opinion on my answers since they are somewhat more familiar with MoveIt.

march_gait_selection/src/march_gait_selection/dynamic_gaits/moveit_subgait.py
```diff
@@ -0,0 +1,66 @@
...
+        rospy.Subscriber('/march/cp_marker_foot_right', Marker, queue_size=1, callback=self.capture_point_cb,
+                         callback_args='right_leg')
+
+    def capture_point_cb(self, msg, leg_name):
+        """Set latest message to variable.
+
+        :param msg: The message from the capture point topic
+        :param leg_name: The name of corresponding move group
+        """
+        self._capture_point_msg[leg_name] = msg
```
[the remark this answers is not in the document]
JorisWeeda: That would be possible but that would require a lot more attributes of the class for both move groups. In my opinion that would not be the best opinion but that is more preference, would u like to see it otherwise?

RutgerVanBeek:
> I have a few questions:
> 
>     * Why are there so many launch files? Do they all have a purpose? Is it correct that only `setup_assistant.launch` is installed?
> 
A lot of them depend on each other, but can also be useful separately. They are all automatically generated. I don't think we need all of them, but some can be useful for testing. The `mach_moveit.launch` is the only one that we will launch from the other parts of the software. This one also depends on other launch files in the directory. 

>     * Why do some files have _fake_ in the name? Should they be pushed?
> 
These are called fake, because they are for testing. I think the near future should show whether we need these. 
>     * Don't you think it is better to include the SRDF file in the `march_description` package? Since it is a description for the march robot and it should probably be modified together with the URDF.
> 

I think this is a good point.

JorisWeeda:
> Don't you think it is better to include the SRDF file in the `march_description` package? Since it is a description for the march robot and it should probably be modified together with the URDF..

@Olavhaasie and @RutgerVanBeek ,  I've thought about this and although it seems like a good idea I do think we should wait with moving the SRDF file. Before we finalize the MoveIt class it could be possible that we are to use the setup assistant again if we want to change the moveIt settings. This would also involve changing the SRDF. I'm pretty sure MoveIt does not handle these sorts of file replacements. We could place the SRDF to another folder after we are (almost) certain that we do not have to use the setup assistant again. would u agree upon this decision?

RutgerVanBeek:
> > Don't you think it is better to include the SRDF file in the `march_description` package? Since it is a description for the march robot and it should probably be modified together with the URDF..
> 
> @Olavhaasie and @RutgerVanBeek , I've thought about this and although it seems like a good idea I do think we should wait with moving the SRDF file. Before we finalize the MoveIt class it could be possible that we are to use the setup assistant again if we want to change the moveIt settings. This would also involve changing the SRDF. I'm pretty sure MoveIt does not handle these sorts of file replacements. We could place the SRDF to another folder after we are (almost) certain that we do not have to use the setup assistant again. would u agree upon this decision?

You are right, I didn't think about this. It is easier if we keep it here for now.

march_gait_selection/src/march_gait_selection/dynamic_gaits/moveit_subgait.py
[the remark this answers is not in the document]
RutgerVanBeek: I thought about this and doubted. You need the time and pose, but not the visualization stuff. On the other hand now you have them both in one attribute.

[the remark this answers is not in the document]
Roelemans: In principle you could just make two variables 
`self._capture_point_time = {'left_leg': 0, 'right_leg, 0}
self._capture_point_position = {'left_leg': [0, 0], 'right_leg': [0,0]}` 
That way you only have one more variable and an intuitive way of getting the relevant info (you also immediately know which info is relevant). i.e. you don't have to look up which fields of the "Marker" msg are relevant. But if you prefer it this way, feel free to keep it like this :)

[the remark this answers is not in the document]
JorisWeeda: hmm I misunderstood your suggestion but this does seem nice, it is indeed more intuitive. I'll make a proper change.

march_gait_selection/src/march_gait_selection/dynamic_gaits/moveit_subgait.py
```diff
@@ -24,8 +24,9 @@ def __init__(self):
             rospy.logerr('Could not connect to move groups, aborting initialisation of the moveit subgait class')
             return
 
-        self._capture_point_msg = {'left_leg': None, 'right_leg': None}
         self._end_effectors = {'left_leg': 'left_foot', 'right_leg': 'right_foot'}
+        self._latest_msg_time = dict()
+        self._poses = dict()
```
Roelemans, line 29:
```suggestion
        self._latest_capture_point_msg_time = {'left_leg': None, 'right_leg': None}
        self._capture_point_position = {'left_leg': None, 'right_leg': None}
```
Or remove "latest_" in first name
Was there a reason you did not initialize these dict keys?

march_gait_selection/src/march_gait_selection/dynamic_gaits/moveit_subgait.py
```diff
@@ -38,7 +39,8 @@ def capture_point_cb(self, msg, leg_name):
         :param msg: The message from the capture point topic
         :param leg_name: The name of corresponding move group
         """
-        self._capture_point_msg[leg_name] = msg
+        self._latest_msg_time[leg_name] = msg.header.stamp
+        self._poses[leg_name] = msg.pose
```
Roelemans, line 43:
```suggestion
        self._latest_capture_point_msg_time[leg_name] = msg.header.stamp
        self._capture_point_position[leg_name] = msg.pose.position
```
You could even consider extracting the x and y coordinate from the msg.pose.position directly and storing vectors in your dict. i.e.
```suggestion
        self._latest_msg_time[leg_name] = msg.header.stamp
        self._poses[leg_name] = [msg.pose.position.x, msg.pose.position.y]
```
Whichever you prefer :)

JorisWeeda: I'm not sure whether storing it like a vector would be beneficial, The move group can use the pose structure like this; 
``` self._move_group[leg_name].set_pose_target(pose, end_effector_link='left_foot')``` 
We can also decide to change it later if necessary and indeed more useful to directly store it like a vector

PR #513, opened 2020-04-21
BLC-26-balance-launch-argument

Ciliaclaij:
Closes BLC-26-balance-launch-argument

## Description

This PR adds a launch argument to the simulation.launch, so balance can be tested in simulation

## Changes

* Added launch argument
* Set default trajectory-execution to false


Ciliaclaij:
I don't really understand what you mean. As far as I understand it, I need to make an parameter in the if statement stat checks the argument balance. But what is the content of this parameter and why are we not able to use the 'balance' parameter?

Olavhaasie:
What I think Rutger is saying, is that you have to upload the balance parameter to the ROS parameter server. Currently, the argument is only used for this launch file and cannot be read from other nodes. Therefore, you need to add a [`<param>`](https://wiki.ros.org/roslaunch/XML/param) tag in the if statement to upload it.

Ciliaclaij:
I just added a parameter in the if statement, but I am not sure if it is in the right place. Maybe we also want the parameter if we don't use balance?

march_launch/launch/march_simulation.launch
```diff
@@ -26,5 +26,6 @@
 
     <group if="$(arg balance)">
         <include file="$(find march_moveit)/launch/march_moveit.launch" />
+        <param name="balance" value="$(arg balance)" />
```
Olavhaasie, line 29: I messed it up. In the first version of my comment I said you should add the `param` to the if statement, but I changed it quickly. So what I meant was to add the `param` tag outside the if statement, since you want to be able to read the `balance` parameter even if it is `false`, because now it will only be uploaded when it is `true`.  I'm sorry :see_no_evil:

Ciliaclaij: Roel just told me haha, I changed it already :)

march_launch/launch/march_simulation.launch
```diff
@@ -18,7 +18,14 @@
...
+    <arg name="balance" default="false" doc="Uses the dynamic balance gaits instead of the predefined gaits."/>
+    <param name="balance" value="$(arg balance)" />
+
     <include file="$(dirname)/march.launch" pass_all_args="true">
         <arg name="configuration" value="simulation"/>
     </include>
+
+    <group if="$(arg balance)">
+        <include file="$(find march_moveit)/launch/march_moveit.launch" />
+    <group>
```
RutgerVanBeek, line 30: I indeed meant what you implemented. When testing I found that you group tags have small syntax error, but they are not really necessary. You can do it as below. 

```suggestion
<include if="$(arg balance)" file="$(find march_moveit)/launch/march_moveit.launch" />
```

march_launch/launch/march_simulation.launch
```diff
@@ -25,7 +25,5 @@
         <arg name="configuration" value="simulation"/>
     </include>
 
-    <group if="$(arg balance)">
-        <include file="$(find march_moveit)/launch/march_moveit.launch" />
-    <group>
+<include if="$(arg balance)" file="$(find march_moveit)/launch/march_moveit.launch" />
```
Olavhaasie, line 28:
```suggestion
    <include if="$(arg balance)" file="$(find march_moveit)/launch/march_moveit.launch" />
```

Ciliaclaij: Lol, I just used the 'Commit suggestion button', which apperently fcked that up. I will try it again

RutgerVanBeek: That means that I made a mistake. Oops.

Olavhaasie: Actually, @RutgerVanBeek messed it up :wink:

PR #515, opened 2020-04-24
Blc 25/add balance gait to state machine

gaiavdh:
Closes BLC-25

## Description
Added the balance walk gait to the state machine. If the balance argument is not used when launching, a warning will be sent when executing the gait. The balance gait will always be created and the Idle state will check if the launch argument is set to true or not. Not creating the balance state machine would cause for a consistency error.

## Changes
* Added the balance walk gait to the state machine.
* launch argument check in the Idle state


JorisWeeda:
It could be possible to not create the balance state machine whenever the balance argument is false but this structure would require to also create a separate standing state for this specific case. Since the balance transition and outcome are not valid. If those would still be added the consistency check of SMACH would fail.

PR #518, opened 2020-05-01
Adjust load_gaits so that default values for walking are imported for…

Deakensbubbles:
Closes BLC-35

## Description

Adjusted the march_gait_selection, in particular the load_gaits() function from gait_selection.py to load the default walk gaits into the balance gait. This allows us to use the right_open subgait from the walk gait. It also makes sure that the gait is filled and will therefore not error the existing setup through the contains_gait() function. The close subgaits have also been added for the same reason.

## Changes

* Adjusted the load_gaits() function
* Adjusted the __getitem__ of Balancegait class

### DIFF
--- a/march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
+++ b/march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
@@ -89,7 +89,5 @@ class BalanceGait(object):
             return self.calculate_trajectory('right_leg')
         elif name == 'left_swing':
             return self.calculate_trajectory('left_leg')
-        elif name in self.default_walk:
-            return self.default_walk[name]
         else:
-            return None
+            return self.default_walk[name]
--- a/march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
+++ b/march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
@@ -85,15 +85,11 @@ class BalanceGait(object):
 
         :param name: the name of the subgait (in this case only left_swing and right_swing should be used)
         """
-        if name == 'right_open':
-            return self.default_walk[name]
-        elif name == 'right_swing':
+        if name == 'right_swing':
             return self.calculate_trajectory('right_leg')
         elif name == 'left_swing':
             return self.calculate_trajectory('left_leg')
-        elif name == 'right_close':
-            return self.default_walk[name]
-        elif name == 'left_close':
+        elif self.default_walk.has_key(name):
             return self.default_walk[name]
         else:
             return None
--- a/march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
+++ b/march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
@@ -89,7 +89,7 @@ class BalanceGait(object):
             return self.calculate_trajectory('right_leg')
         elif name == 'left_swing':
             return self.calculate_trajectory('left_leg')
-        elif self.default_walk.has_key(name):
+        elif name in self.default_walk:
             return self.default_walk[name]
         else:
             return None
--- a/march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
+++ b/march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
@@ -12,6 +12,7 @@ class BalanceGait(object):
 
     def __init__(self, gait_name='gait_balanced_walk'):
         self.gait_name = gait_name
+        self.default_walk = None
         self._is_balance_used = rospy.get_param('/balance', False)
 
         if self._is_balance_used:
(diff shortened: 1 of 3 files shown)


march_gait_selection/src/march_gait_selection/gait_selection.py
```diff
@@ -72,6 +71,8 @@ def load_gaits(self):
             loaded_gait = Gait.from_file(gait, self.gait_directory, self.robot, self._gait_version_map)
             self.loaded_gaits.append(loaded_gait)
 
+        self.balance_gait.default_walk = next((gait for gait in self.loaded_gaits if gait.gait_name == 'walk'), None)
```
Roelemans, line 74:
```suggestion
        self.balance_gait.default_walk = self['walk']
```
This uses the __getitem__ method ;)

JorisWeeda:
> I was busy with the capture point based subgaits and the loaded_gaits contains gait object which return subgait objects when you use the getitem. We are returning a subgait message now, but this should actually be a Subgait object.
> 
> I think there are two possible solutions:
> 
>     1. Store the subgait msg in a gait object. This means a lot of refactoring of the gait selection I am not so familiar with that.
> 
>     2. Import the Gait/Subgait module and create a subgait object from the subgait msg and return that.
> 
> 
> EDIT: I see that there is no `to_subgait_msg` function so that one should be created first.
> 
> EDIT 2: There is also a less nice fix, by checking whether it is a balanced swing gait and then just returning the msg instead of the conversion from object to message.

This is indeed an important detail. The transition step creates a subgait object when used so maybe to be consistent we should use that structure here as well. If we decide to do otherwise I'll change it within the transition gait as well.

RutgerVanBeek:
I have pushed my solution. This is not ideal, but better solutions require a big refactor. I think we might consider to make a gait abstract base class, which StaticGait(Current gait) and BalanceGait can inherit from for a clear interface for the gait selection and rqt gait selection.

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
```diff
@@ -84,9 +85,15 @@ def __getitem__(self, name):
...
+        if name == 'right_open':
+            return self.default_walk[name]
+        elif name == 'right_swing':
             return self.calculate_trajectory('right_leg')
         elif name == 'left_swing':
             return self.calculate_trajectory('left_leg')
+        elif name == 'right_close':
+            return self.default_walk[name]
+        elif name == 'left_close':
+            return self.default_walk[name]
```
Roelemans, line 97:
```suggestion
        if name == 'right_swing':
            return self.calculate_trajectory('right_leg')
        elif name == 'left_swing':
            return self.calculate_trajectory('left_leg')
        elif self.default_walk.has_key(name):
            return self.default_walk[name]
```

RutgerVanBeek:
> @JorisWeeda and I actually considered having BalancedGait inherit from the Gait class. One reason why we did not do this is because it would cause multiple initializations of MoveIt when one has mutliple balanced gaits. We also considered having BalancedGait inherit from Subgait and wrapping it in a gait instance, but that too had similar problems. I suspect a good solution would have mulitple balanced gait/subgait instances, which all contain a reference to the same, unique, MoveIt interface object.
> 
> I think we should currently just keep things a bit "beuned" as it is now. We can implement this later if/when we actually get this to work.

Yes, I agree that this is something for the future, when we get everything working and have better idea about the implementation.

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
```diff
@@ -88,5 +89,7 @@ def __getitem__(self, name):
             return self.calculate_trajectory('right_leg')
         elif name == 'left_swing':
             return self.calculate_trajectory('left_leg')
+        elif name in self.default_walk:
```
RutgerVanBeek, line 92: It gets stuck in that line, I don't understand why yet.

JorisWeeda: As of now we do understand why. In order to use the _ for item in self_ u must define a **container**. However the docs do state that if the **container** is not specified it uses the getitem. But since the program freezes in our case that does not seem to work. I tested this defining a container in the gait class and this seems to work properly. Look at; https://docs.python.org/3/reference/datamodel.html?emulating-container-types#emulating-container-types for more information

Roelemans: My mistake, I for some reason assumed default_walk was a dictionary, but it is a gait class instead.

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
```diff
@@ -88,5 +89,7 @@ def __getitem__(self, name):
...
+            return self.default_walk[name]
         else:
             return None
```
Roelemans, line 95: You can simply do
```suggestion
        else:
            return self.default_walk[name]
```
(self.default_walk[name] returns None if name not in self.default_walk)

JorisWeeda: simple yet elegant, good solution.

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
JorisWeeda: Maybe in the future it would be wise if we invest in defining more of the data type function so we could use the class like a dictionary

PR #519, opened 2020-05-01
Fix namespace issue when launching the moveit package. 

JorisWeeda:
## Description
There was a namespace issue because the moveit package was not included within the march namespace inside the march.launch. It is also more logical to include a launch file in the march.launch since all the other launch files are also located here.  

## Changes
* Moved the march_moveit launch to the march.launch file
* removed the march_moveit launch from the march_simulation.launch file

### DIFF
--- a/march_launch/launch/march.launch
+++ b/march_launch/launch/march.launch
@@ -48,7 +48,7 @@
         <include file="$(find march_state_machine)/launch/state_machine.launch">
             <arg name="state_machine_viewer" value="$(eval not headless and state_machine_viewer)"/>
             <arg name="unpause" value="$(eval configuration == 'simulation' and unpause)"/>
-            <arg name="sounds" value="$(arg sounds)"/>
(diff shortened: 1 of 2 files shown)


march_launch/launch/march.launch
```diff
@@ -48,7 +48,7 @@
         <include file="$(find march_state_machine)/launch/state_machine.launch">
             <arg name="state_machine_viewer" value="$(eval not headless and state_machine_viewer)"/>
             <arg name="unpause" value="$(eval configuration == 'simulation' and unpause)"/>
-            <arg name="sounds" value="$(arg sounds)"/>
+            <arg name="sounds" value="$(arg sounds)"/>
```
Roelemans, line 51: Whats this

JorisWeeda: i don't know, probably some tabs

JorisWeeda: git does not recognize these spaces so maybe a bug or something?

PR #522, opened 2020-05-14
extend gain scheduling timeout

RutgerVanBeek:
Closes PM-428

## Description

With the implementation of the writing of iMC projects to the drive via Ethercat startup can be quite long when this need to happen. It was found that this timeout caused exo to restarted, because gain scheduling only waits 30 seconds for the controllers come up, which only happens when the joints are initialized well. I think about 45 seconds are needed to write the projects to the drive so I made it double of that to be sure. 
## Changes

Extend the timeout for the gain scheduling node to wait for the controller topics from 30 to 90 seconds.


RutgerVanBeek:
> Why did you push so many log files? I guess this is a fault in the commit

Yes, I fucked up.

Deakensbubbles:
This seems like a nice change. And because the startup time isn't crucial, the 90 seconds seems like a good educated guess if the maximum startup time seems to be around 45 seconds.

JorisWeeda:
Same problem in Travis with a certain catkin package link, restarting the job previously worked so I'll start it again and let's hope for the best!

PR #525, opened 2020-05-22
Feature: add alive msg

Olavhaasie:
First step of PM-446

## Description
This PR adds an `Alive` msg, which contains a timestamp and identifier. Previously, there existed no such message and `Time` was just used. I added a comment, which describes what an id should be. I'm open to suggestions about what this field should contain.

I have also looked into other ways of identifying input devices. Using the node name seemed like a good idea, since these are always unique and no extra messages are necessary. However, the way in which you would have to retrieve this name is [_brittle_](https://answers.ros.org/question/10346/how-to-get-the-publisher-name-from-a-received-message-in-rospy/?answer=15240#post-id-15240).

Personally, I find the `Alive` message a nice way, since you can directly see who and what sent the message, also programmatically.

## Changes
* Added `Alive.msg`
* Added `id` to `GaitInstruction.msg`

### DIFF
--- a/march_shared_resources/CMakeLists.txt
+++ b/march_shared_resources/CMakeLists.txt
@@ -13,6 +13,7 @@ find_package(catkin REQUIRED COMPONENTS
 add_message_files(
     FILES
     AfterLimitJointCommand.msg
+    Alive.msg
     Error.msg
     GaitInstruction.msg
     GaitInstructionResponse.msg
--- a/march_shared_resources/msg/Alive.msg
+++ b/march_shared_resources/msg/Alive.msg
@@ -1,0 +1,9 @@
+# Message type that is used by input devices to signal they are still alive.
+
+# Timestamp at which the input device was last alive
+time stamp
+
+# Unique identifier used to identify an input device when connection has been lost.
+# The id should be readable and you should be able to tell from what machine and
+# what type of input device it originates.
+string id
--- a/march_shared_resources/msg/GaitInstruction.msg
+++ b/march_shared_resources/msg/GaitInstruction.msg
@@ -15,3 +15,5 @@ int8 DECREMENT_STEP_SIZE = 5
 # If the type is a GAIT this is the name of the gait.
 string gait_name
 
+# Unique identifier for the input device sending this message. See `Alive`.
+string id


RutgerVanBeek:
> Good change. Not sure what kind of ID would be good. You might actually want to be able to differentiate different laptops ssh'ing and starting a developer ipd, so that you can blame someone if a wrong button was pressed . That would support using something like their username, but that is not unique :/

May be you should enter an ID's as a user, for the real IPD this can be something like realIPD. When I am in my terminal I see my computername. i.e. `rutger-HP-ProBook-440-G3`, can you acces that? (something like `socket.gethostname()`).

Olavhaasie:
> May be you should enter an ID's as a user, for the real IPD this can be something like realIPD. When I am in my terminal I see my computername. i.e. `rutger-HP-ProBook-440-G3`, can you acces that? (something like `socket.gethostname()`).

Yes that was what I had in mind. I am experimenting with it now in the rqt input device and I have come up with `rqt{serial}@{machine}@{user}`. The serial is the serial id of the input device in rqt, since you can have multiple open at the same time. This is probably not used in the real world, but is useful for testing :sweat_smile:. The other two are self-explanatory.

And for the embedded IPD we could do something similar `crutch@{processor}` :shrug: Not so sure, but haven't thought about the real IPD yet. Perhaps add the version.

RutgerVanBeek:
> > May be you should enter an ID's as a user, for the real IPD this can be something like realIPD. When I am in my terminal I see my computername. i.e. `rutger-HP-ProBook-440-G3`, can you acces that? (something like `socket.gethostname()`).
> 
> Yes that was what I had in mind. I am experimenting with it now in the rqt input device and I have come up with `rqt{serial}@{machine}@{user}`. The serial is the serial id of the input device in rqt, since you can have multiple open at the same time. This is probably not used in the real world, but is useful for testing . The other two are self-explanatory.
> 
> And for the embedded IPD we could do something similar `crutch@{processor}`  Not so sure, but haven't thought about the real IPD yet. Perhaps add the version.

This seems like a very good option. Why should the real IPD have the same structure, I think it should mostly be readable and identify the user. I do not see the benefit of having several ID's for several IPD's on the same machine, because one would still not know which was which IPD.

Olavhaasie:
> I do not see the benefit of having several ID's for several IPD's on the same machine, because one would still not know which was which IPD.

True, you could always add specific names if you ever have multiple IPDs. For example, if you have devices in both crutches, you could id them `left_crutch` and `right_crutch`

PR #526, opened 2020-05-26
Set the load robot description argument to default false

JorisWeeda:
## Description
When using the moveit package (with the balance argument set to true) the robot_description parameter was set by default in the moveit launch file. In combination with the simulation model a problem occurred. The march_world.launch sets a few variables to the robot description before setting the parameter to the ROS server. Since moveit launched the parameter in his own launch file the changes set by the march simulation launch were not received in the moveit package. 

By setting the default value of load_robot_description in the moveit launch file to false the robot_description used in the simulation model is also used in the moveit package.

### DIFF
--- a/march_moveit/launch/move_group.launch
+++ b/march_moveit/launch/move_group.launch
@@ -37,7 +37,7 @@
                 " />
   -->
 
-  <arg name="load_robot_description" default="true" />
+  <arg name="load_robot_description" default="false" />
   <!-- load URDF, SRDF and joint_limits configuration -->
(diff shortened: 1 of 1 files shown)


RutgerVanBeek:
Shouldn't it also be removed from the march_launch package to make sure there is no double upload?

Olavhaasie:
> Shouldn't it also be removed from the march_launch package to make sure there is no double upload?

Only the parameter that is [last set](https://wiki.ros.org/roslaunch/XML#Evaluation_order) will be preserved. So the robot description in march_launch will be uploaded first and then overridden by this one.

PR #529, opened 2020-05-29
Feature/blc 32 subgait based on capture point

JorisWeeda:
Closes BLC-32

## Description
This PR creates the possibility to create a subgait which uses the capture point of the left or right leg during the right or left swing. It combines the capture point trajectory with the default subgait in order to create a movement for all the joints in the exoskeleton.

Currently there are a few function in this class which are also declared in the transition gait and probably also the gait generator. In a separate issue I will migrate these functions to the subgait class and update all the corresponding dependencies.  

## Changes
* Constructing a subgait with the movegroup joints based on the capture point trajectory
* Merged the move group subgait with the default subgait
* Scheduled the capture point subgait during the left and right swing


march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
```diff
@@ -81,14 +103,117 @@ def to_subgait_msg_(name, trajectory_msg, gait_type='walk_like', version='moveit
...
+            balance_joints.append(JointTrajectory(joint_name, normal_joint.limits, setpoints, balance_duration))
+
+        balance_subgait = BalanceGait.to_subgait(balance_joints, balance_duration)
+
+        return balance_subgait
+
+    @staticmethod
+    def _equalize_amount_of_setpoints(first_subgait, second_subgait):
+        """Equalize the subgaits to have matching amount of setpoints on all the timestamps."""
+        max_duration = max([first_subgait.duration, second_subgait.duration])
```
RutgerVanBeek, line 136:
```suggestion
        max_duration = max(first_subgait.duration, second_subgait.duration)
```

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
```diff
@@ -44,34 +51,49 @@ def capture_point_cb(self, msg, leg_name):
...
+
+    def random_subgait(self):
+        """Create random trajectory using the moveit motion planner.
+
+        :return:
+            A populated subgait message
+        """
+        self._move_group['left_leg'].set_random_target()
+        trajectory_msg = self._move_group['left_leg'].plan()
+        return self.to_subgait_msg('Random moveit subgait', trajectory_msg.joint_trajectory)
```
RutgerVanBeek, line 64: If we want to keep this one should plan for both move groups and merge the result.

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
```diff
@@ -81,14 +103,117 @@ def to_subgait_msg_(name, trajectory_msg, gait_type='walk_like', version='moveit
...
+        all_timestamps = first_subgait.get_unique_timestamps() + second_subgait.get_unique_timestamps()
+        return sorted(set([round(timestamp, Setpoint.digits) for timestamp in all_timestamps]))
+
+    @staticmethod
+    def merge_subgaits(balance_trajectory_subgait, subgait):
+        balance_trajectory_subgait, subgait = \
+            BalanceGait._equalize_amount_of_setpoints(balance_trajectory_subgait, subgait)
+
+        if subgait.duration != balance_trajectory_subgait.duration:
+            return None
```
RutgerVanBeek, line 188: May be log a warning or error here

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
JorisWeeda: we could just delete this class. I do not expect we will use this function in any situation.

RutgerVanBeek: yes, was nice for testing but does not seem necessary anymore

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
```diff
@@ -81,14 +103,140 @@ def to_subgait_msg_(name, trajectory_msg, gait_type='walk_like', version='moveit
...
+    @staticmethod
+    def create_subgait_of_trajectory(normal_subgait, joint_trajectory):
+        """Create a subgait using the joint trajectory generated by the capture point pose.
+
+        :param normal_subgait: The normal subgait to get the joint limits
+        :param joint_trajectory: The capture point pose trajectory
+
+        :return: a populated subgait object generated from the joint trajectory
+        """
+        balance_duration = joint_trajectory.points[-1].time_from_start.to_sec()
```
RutgerVanBeek, line 121: I got a list index out of range error here. This means that the trajectory message has no points. This happens because no motion plan is found. (It was lying on the ground) I think you should check whether the trajectory message has a non-zero amount of points, because apparently what MoveIt does when no motion plan is found is not throw an error or None, but return an empty trajectory msg.

JorisWeeda: done

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
```diff
@@ -28,9 +33,11 @@ def __init__(self, gait_name='gait_balanced_walk'):
                 rospy.logerr('Could not connect to move groups, aborting initialisation of the moveit subgait class')
                 return
 
-        self._end_effectors = {'left_leg': 'left_foot', 'right_leg': 'right_foot'}
+        self._end_effectors = {'left_leg': 'foot_left', 'right_leg': 'foot_right'}
```
Olavhaasie, line 36: **left**\_leg, but foot\_**left**? I think this was better before, this could cause a lot of confusion and nasty bugs

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
```diff
@@ -44,34 +51,39 @@ def capture_point_cb(self, msg, leg_name):
...
-            A populated subgait message
-        """
-        self._move_group['left_leg'].set_random_target()
-        trajectory_msg = self._move_group['left_leg'].plan()
-        return self.to_subgait_msg_('Random moveit subgait', trajectory_msg.joint_trajectory)
+        self._old_capture_point_msg_time[leg_name] = self._latest_capture_point_msg_time[leg_name]
+
+        _pose = self._capture_point_pose[leg_name]
+        _end_effector = self._end_effectors[leg_name]
+        self._move_group[leg_name].set_joint_value_target(_pose, _end_effector, True)
```
Olavhaasie, line 77: local variables mustn't use a preceding underscore. This confused me, because I thought you were missing `self.`
```suggestion
        pose = self._capture_point_pose[leg_name]
        end_effector = self._end_effectors[leg_name]
        self._move_group[leg_name].set_joint_value_target(pose, end_effector, True)
```

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
```diff
@@ -81,14 +93,146 @@ def to_subgait_msg_(name, trajectory_msg, gait_type='walk_like', version='moveit
...
+            joint.setpoints[-1].time = new_duration
+
+        subgait.duration = new_duration
+        return subgait
+
+    @staticmethod
+    def _get_all_unique_timestamps(first_subgait, second_subgait):
+        """Get all the timestamps from the subgaits, eliminate double."""
+        all_timestamps = first_subgait.get_unique_timestamps() + second_subgait.get_unique_timestamps()
+        return sorted(set([round(timestamp, Setpoint.digits) for timestamp in all_timestamps]))
```
Olavhaasie, line 177: Why is this method placed in the `BalanceGait` class? This method can be used for every type of gait with subgaits. In fact, a lot of other methods can also be moved from this class, in order to keep this class simple: `_scale_timestamps_subgaits`, `_equalize_amount_of_setpoints`, `create_subgait_of_trajectory` and `merge_subgaits`

march_moveit/CMakeLists.txt
```diff
@@ -1,9 +1,9 @@
 cmake_minimum_required(VERSION 3.0.2)
 project(march_moveit)
 
-find_package(catkin REQUIRED)
+find_package(catkin REQUIRED COMPONENTS march_description)
 
-catkin_package()
+catkin_package(CATKIN_DEPENDS march_description)
```
Olavhaasie, line 6: I removed this dependency since it was not necessary for building the package. Why did you add it again?

march_moveit/package.xml
```diff
@@ -17,11 +17,12 @@
 
   <buildtool_depend>catkin</buildtool_depend>
 
+  <depend>march_description</depend>
```
Olavhaasie, line 20: I found that this was only necessary as `exec_depend`, since you only upload the urdf when you launch moveit. Why did you make it `depend`?

march_state_machine/src/march_state_machine/state_machines/gait_state_machine.py
```diff
@@ -59,15 +60,18 @@ def add_subgait(self, subgait_name, succeeded='succeeded', stopped=None):
...
+                subgaits = self.get_children().keys()
+                if not self._contains_gait(gait=self._gait_name, subgaits=subgaits).contains:
+                    rospy.logwarn(
+                        'Gait {0} is not currently loaded with subgaits {1}'.format(self._gait_name, subgaits))
+                    return 'rejected'
+            except rospy.ServiceException:
+                rospy.logerr(
+                    'Failed to contact {0}, is a gait_selection node running?'.format(
+                        self._contains_gait.resolved_name))
+                return 'failed'
```
Olavhaasie, line 74: Why do you skip this check? Why not just let the gait selection check for the subgaits? The balance gait should be checked as well right? this seems like a cheap solution

march_moveit/package.xml
JorisWeeda: Got some [troubles with Travis](https://travis-ci.com/github/project-march/march/jobs/341377861) and thought this would be a fitting solution

march_moveit/CMakeLists.txt
JorisWeeda: See comment on the package.xml

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
JorisWeeda: This sprint I have an issue which migrates these functionalities to the subgait class and refactor the balance/transition gait and the gait generator

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
JorisWeeda: Indeed not the best, did not think this through

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
JorisWeeda: I will fix this

march_moveit/package.xml
Olavhaasie: You received that error because it was declared a `depend` in the `package.xml`, if you revert both of these files it should be fine. That is what worked in my PR #531

march_state_machine/src/march_state_machine/state_machines/gait_state_machine.py
Olavhaasie: You could make the check for the balance gait succeed when the names exist?

JorisWeeda: That is originally what the function does, it checks if **gait[subgait]** returns a _not None_ value. But doing this operation constructs a subgait for the balance gait as the ```__getitem__``` method states. I'm not sure if there is a better solution except for changing the balance gait ```__getitem__``` method

march_moveit/package.xml
JorisWeeda: Nice, tbh I was fairly confused by the travis build failing.

march_moveit/launch/planning_context.launch
```diff
@@ -3,13 +3,13 @@
   <arg name="load_robot_description" default="false"/>
 
   <!-- The name of the parameter under which the URDF is loaded -->
-  <arg name="robot_description" default="/robot_description"/>
```
RutgerVanBeek: This was an implicit remap, which is necessary. Either revert this or make an explicit remap of the variable.

JorisWeeda: The ```/``` is necessary because otherwise it will create a duplicate robot model which does not constrain itself to the variables set in the _march_world.launch_. Adding the ```/``` will make sure every part of code uses the same robot_description parameter from the ROS server.

RutgerVanBeek: I don't fully understand, because when I added this comment, GitHub said you removed the `/` and also when I pull, I don't see it, but I agree that it is necessary.

JorisWeeda: it seems like I screwed up my last commit with the setup_assistant again, my bad.

PR #533, opened 2020-06-04
Feature esp stateful windows

RutgerVanBeek:
Closes PM-407

## Description

The idea was to have a csv output for stateful windows via a ROS shutdown hook. I found another way via a RemoveState window, which turns stateful windows into stateless windows. Next to that I put I fixed that inserting events with a similar time was not possible. There is retention throughout the model(This means that old events are deleted for better performance). Finally I also added the new data from the 'ImcState.msg', the velocities form the encoders and the iMC voltage.
## Changes

* Added RemoveState windows for csv output of statefull windows
* Added retention throughout the model
* Added the new variables in the `ImcState.msg`


march_data_collector/src/march_data_collector/esp_adapter.py
```diff
@@ -271,23 +279,37 @@ def pid_state_callback(self, data, sources):
...
         join_time_str = get_join_time_str(data.header.stamp, self.control_analysis_join_frequency)
+
+        if self.previous_join_key[sources[0]] == join_time_str:
+            return
+        self.previous_join_key[sources[0]] = join_time_str
+
         csv = list_to_str([data.p_error, data.i_error, data.d_error, data.p_term, data.i_term,
                           data.d_term, data.output])
-        self.send_to_esp('{0}, {1}, {2}'.format(time_str, join_time_str, csv), sources)
+        self.send_to_esp('{0}, {1}, {2}, 1'.format(time_str, join_time_str, csv), sources)
```
Olavhaasie, line 289: What does the `1` stand for?

RutgerVanBeek:
> > Added the new variables in the ImcState.msg
> 
> I don't see this change?

What I meant is that added the new variables in the msg to ESP.

> I was also thinking about the `ImcState.msg`, we might have to split the message into multiple messages, if we keep adding new variables. It would be nice if the message only contains the state as in: `Operational`, `Fault`, etc. with error descriptions. The velocities, currents and encoders should be in their own message. This way we don't have to keep changing this message every time we want something from the IMC. I know it is not really relevant to this PR, but something to think about

Yes, or we rename it to `ImcData.msg` and then only publish the state if there is something special.

march_data_collector/src/march_data_collector/esp_adapter.py
RutgerVanBeek: This is a dummy key, which I use in more event streams. Sometimes you want to take things like means or maxima and you do that over a group of events, but they should have something in common. That something is a dummy key here and I have decided that dummy keys should be an integer one.

march_data_collector/src/march_data_collector/esp_adapter.py
```diff
@@ -339,7 +361,7 @@ def get_join_time_str(timestamp, frequency):
 
     :param data: ROS timestamp message std_msgs/stamp
     """
-    time = round(timestamp.secs + timestamp.nsecs * 10 ** (-9) * frequency) / frequency
+    time = round((timestamp.secs + timestamp.nsecs * 10 ** (-9)) * frequency) / frequency
```
JorisWeeda, line 364: I think this should also work, not extremely relevant but a nice touch maybe. 
```suggestion
    time = round(timestamp.to_sec() * frequency) / frequency
```

Olavhaasie: @RutgerVanBeek are you implementing this change or not? Since you gave a :+1:

RutgerVanBeek: Yes I am planning on doing this today.

PR #538, opened 2020-06-18
Efficient implemention of setpoint interpolation

Roelemans:
Closes PM-478
Merge together with https://github.com/project-march/gait-generation/pull/99
The built fails because of this dependency :s

## Description

For some reason, every time an interpolated setpoint was requested the following happened: 1) the entire gait would be interpolated 2) the interpolation would be sampled every 0.01 seconds 3) the two samples closest to the requested time would be found 4) these would be interpolated linearly to get the result. My eyes starting tearing by just looking at it.

In this PR, the interpolation is done only once at startup. The interpolation is saved as a function rather than a list of samples. Life is better now.

Unfortunately, startup is not faster as I once announced. Because the interpolation was never done at startup, only every time any interpolated point was requested. Since the interpolation is now done at startup, it is actually a bit slower instead. Loading the gait selection takes on average:
Before: 4.29 seconds
After: 4.91 seconds

## Changes

* Do interpolation during initialization of a JointTrajectory object
* Save interpolation as two functions instead of a list of interpolated points

### DIFF
--- a/march_shared_classes/src/march_shared_classes/gait/joint_trajectory.py
+++ b/march_shared_classes/src/march_shared_classes/gait/joint_trajectory.py
@@ -12,9 +12,9 @@ class JointTrajectory(object):
     def __init__(self, name, limits, setpoints, duration, *args):
         self.name = name
         self.limits = limits
-        self.setpoints = setpoints
-        self.duration = duration
-        [self.interpolated_position, self.interpolated_velocity] = self.interpolate_setpoints()
+        self._setpoints = setpoints
+        self._duration = duration
+        self.interpolate_setpoints()
 
     @classmethod
     def from_dict(cls, subgait_dict, joint_name, limits, duration, *args):
@@ -39,6 +39,24 @@ class JointTrajectory(object):
                                                 point['velocities'][joint_index]))
 
         return cls(joint_name, limits, setpoints, duration, *args)
+
+    @property
+    def duration(self):
+        return self._duration
+
+    @duration.setter
+    def duration(self, duration):
+        self._duration = duration
+        self.interpolate_setpoints()
+
+    @property
+    def setpoints(self):
+        return self._setpoints
+
+    @setpoints.setter
+    def setpoints(self, setpoints):
+        self._setpoints = setpoints
+        self.interpolate_setpoints()
 
     def get_setpoints_unzipped(self):
         """Get all the listed attributes of the setpoints."""
@@ -93,9 +111,8 @@ class JointTrajectory(object):
 
         # We do a cubic spline here, just like the ros joint_trajectory_action_controller,
         # see https://wiki.ros.org/robot_mechanism_controllers/JointTrajectoryActionController
-        position = BPoly.from_derivatives(time, yi)
-        velocity = position.derivative()
-        return [position, velocity]
+        self.interpolated_position = BPoly.from_derivatives(time, yi)
+        self.interpolated_velocity = self.interpolated_position .derivative()
 
     def get_interpolated_setpoint(self, time):
         if time < 0:
--- a/march_shared_classes/src/march_shared_classes/gait/joint_trajectory.py
+++ b/march_shared_classes/src/march_shared_classes/gait/joint_trajectory.py
@@ -14,6 +14,8 @@ class JointTrajectory(object):
         self.limits = limits
         self._setpoints = setpoints
         self._duration = duration
+        self.interpolated_position = None
+        self.interpolated_velocity = None
         self.interpolate_setpoints()
 
     @classmethod
@@ -102,7 +104,9 @@ class JointTrajectory(object):
 
     def interpolate_setpoints(self):
         if len(self.setpoints) == 1:
-            return [lambda time: self.setpoints[0].position, lambda time: self.setpoints[0].velocity]
+            self.interpolated_position = lambda time: self.setpoints[0].position
+            self.interpolated_velocity = lambda time: self.setpoints[0].velocity
+            return
 
         time, position, velocity = self.get_setpoints_unzipped()
         yi = []
@@ -112,7 +116,7 @@ class JointTrajectory(object):
         # We do a cubic spline here, just like the ros joint_trajectory_action_controller,
         # see https://wiki.ros.org/robot_mechanism_controllers/JointTrajectoryActionController
         self.interpolated_position = BPoly.from_derivatives(time, yi)
-        self.interpolated_velocity = self.interpolated_position .derivative()
+        self.interpolated_velocity = self.interpolated_position.derivative()
 
     def get_interpolated_setpoint(self, time):
         if time < 0:
--- a/march_shared_classes/src/march_shared_classes/gait/joint_trajectory.py
+++ b/march_shared_classes/src/march_shared_classes/gait/joint_trajectory.py
@@ -1,4 +1,3 @@
-import numpy as np
 import rospy
 from scipy.interpolate import BPoly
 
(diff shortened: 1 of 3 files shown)


march_shared_classes/src/march_shared_classes/gait/joint_trajectory.py
```diff
@@ -15,6 +14,7 @@ def __init__(self, name, limits, setpoints, duration, *args):
         self.limits = limits
         self.setpoints = setpoints
         self.duration = duration
+        [self.interpolated_position, self.interpolated_velocity] = self.interpolate_setpoints()
```
JorisWeeda, line 17: wouldn't it be better to populate these variables inside the`interpolate_setpoints(self)`. if (for example) the duration of the subgait, and therefore the joint, is changed the function has to be called again and `self.interpolated_position` and `self.interpolated_velocity` have to be recalculated. This seems like something to do inside the class. 

Another solution could be to define the self.duration as a property and recalculate the interpolation whenever the duration is changed.

Roelemans: Well spotted :) I improved it as suggested

march_shared_classes/src/march_shared_classes/gait/joint_trajectory.py
```diff
@@ -83,32 +101,28 @@ def _validate_boundary_points(self):
...
             yi.append([position[i], velocity[i]])
 
         # We do a cubic spline here, just like the ros joint_trajectory_action_controller,
         # see https://wiki.ros.org/robot_mechanism_controllers/JointTrajectoryActionController
-        position = BPoly.from_derivatives(time, yi)
-        velocity = position.derivative()
-        indices = np.linspace(0, self.duration, self.duration * 100)
-        return [indices, position(indices), velocity(indices)]
+        self.interpolated_position = BPoly.from_derivatives(time, yi)
+        self.interpolated_velocity = self.interpolated_position .derivative()
```
Olavhaasie, line 115:
```suggestion
        self.interpolated_velocity = self.interpolated_position.derivative()
```

march_shared_classes/src/march_shared_classes/gait/joint_trajectory.py
```diff
@@ -83,32 +101,28 @@ def _validate_boundary_points(self):
                (self.setpoints[-1].time == round(self.duration, Setpoint.digits) or self.setpoints[-1].velocity == 0)
 
     def interpolate_setpoints(self):
+        if len(self.setpoints) == 1:
+            return [lambda time: self.setpoints[0].position, lambda time: self.setpoints[0].velocity]
```
Olavhaasie, line 105: Why do you return lambdas here in a method that does not actually return anything?

march_shared_classes/src/march_shared_classes/gait/joint_trajectory.py
Olavhaasie, line 115: You initialize the `self` variables here outside of the `__init__` method, which means they will only be available after calling this method. For clarity it is better to initialize these variables in the `__init__` so that they can be accessed, or better yet make `interpolate_setpoints` return the values so that you can set them just like you did with the lambdas from the comment above.

march_shared_classes/src/march_shared_classes/gait/joint_trajectory.py
```diff
@@ -83,32 +101,28 @@ def _validate_boundary_points(self):
...
-                velocity = interpolated_setpoints[2][i - 1]
-                return self.setpoint_class(time, position, velocity)
-        rospy.logerr('Could not interpolate setpoint at time {0}'.format(time))
-        return self.setpoint_class(0, 0, 0)
+        if time < 0:
+            rospy.logerr('Could not interpolate setpoint at time {0}'.format(time))
+            return self.setpoint_class(time, self.setpoints[0].position, 0)
+        if time > self.duration:
+            rospy.logerr('Could not interpolate setpoint at time {0}'.format(time))
+            return self.setpoint_class(time, self.setpoints[-1].position, 0)
```
Olavhaasie, line 123: This should be a warning message I think, since you can continue with your normal flow only you had to make some changes. I think an error should be logged when something could not complete.

Roelemans: Hmm an error is not a fatal though, and this can only happen if there is a bug in the code. I think you should shut down and track the error before continuing. (though fatal would not be necessary)

march_shared_classes/src/march_shared_classes/gait/joint_trajectory.py
Roelemans: Cus I am stoopid :cry:

march_shared_classes/src/march_shared_classes/gait/joint_trajectory.py
Roelemans: I do call that method in the init though :/ seems a bit redundant to intialize these to None right before the method in which they get set to a useful value. I actually had the implementation where it returns these values before Joris suggested this format. What do you think is best?

march_shared_classes/src/march_shared_classes/gait/joint_trajectory.py
```diff
@@ -40,6 +42,24 @@ def from_dict(cls, subgait_dict, joint_name, limits, duration, *args):
 
         return cls(joint_name, limits, setpoints, duration, *args)
 
+    @property
+    def duration(self):
+        return self._duration
+
+    @duration.setter
+    def duration(self, duration):
+        self._duration = duration
```
JorisWeeda, line 51: if the duration is set the timestamps of the individual setpoints have to be changed as well. In the current situation this is not the case and therefore the test (below) fails.

```    
def test_set_duration(self):
        new_duration = self.joint_trajectory.duration / 2
        self.joint_trajectory.duration = new_duration
        self.assertTrue(all(setpoint.time <= new_duration for setpoint in self.joint_trajectory.setpoints))
```

Roelemans: Its not really a bug, since the duration is only changed inside undo, redo and Subgait.scale_timestamps_subgait, and the setpoints are changed first in these cases. Hence, it is at most a design flaw. I think the solution would be to remove this setter, and use a set_duration method instead, where we can input a "rescale" parameter. That way, it is no longer necessary to directly change setpoints in Subgait.scale_timestamps_subgait, as that will be done in JointTrajectory.set_duration. What do you think?

JorisWeeda: I agree with the set_duration method. My original thought was looking at the code the ```duration.setter``` could be used stand alone which would screw up the joint trajectory (not that we use it like that but it is implied). Removing this property would be wise. We could even use this method in the subgait instead of looping over every setpoint.

Roelemans: Okay

march_shared_classes/src/march_shared_classes/gait/joint_trajectory.py
```diff
@@ -40,6 +42,30 @@ def from_dict(cls, subgait_dict, joint_name, limits, duration, *args):
 
         return cls(joint_name, limits, setpoints, duration, *args)
 
+    @property
+    def duration(self):
+        return self._duration
```
JorisWeeda, line 47: I guess this is kind of redundant right? maybe just change the ```self._duration``` to ```self.duration```?

Roelemans: I chose this because I want to make explicit that self.duration should not be modified directly

JorisWeeda: allright.

PR #543, opened 2020-06-23
Feature/blc 42 smooth capture point

RutgerVanBeek:
Closes BLC-42

## Description

As also stated in https://github.com/project-march/simulation/pull/57, capture point was too noisy. One of the underlying reasons was the updating of TF, but another reasons is that the world is a messy place. Filtering the signal is therefore a good idea. I have chosen a Savitzky-Golay filter (https://en.wikipedia.org/wiki/Savitzky%E2%80%93Golay_filter), which is seems to be a good filter when you need to differentiate. It fits a certain order polynomial to the last x points of your data and thats your new signal. 

## Changes

* Store CoM's in a Ringbuffer.
* Differentiate the CoM position with the Savitzky-Golay algorithm form the scipy package.

### DIFF
--- a/march_data_collector/src/march_data_collector/cp_calculator.py
+++ b/march_data_collector/src/march_data_collector/cp_calculator.py
@@ -30,7 +30,7 @@ class CPCalculator(object):
 
         self.g = 9.81  # gravity constant
         self.buffer_size = 25  # number of com's to use in estimation of velocity
-        self.polyorder = 4  # order of the polynomial in Saviztky-Golay
+        self.polyorder = 4  # order of the polynomial in Savitzky-Golay
 
         self.com_x_buffer = RingBuffer(capacity=self.buffer_size, dtype=float64)
         self.com_y_buffer = RingBuffer(capacity=self.buffer_size, dtype=float64)
@@ -38,18 +38,17 @@ class CPCalculator(object):
     def calculate_cp(self, com_mark):
         current_time = com_mark.header.stamp
         if current_time is not self.prev_t:
-
             self.com_x_buffer.append(com_mark.pose.position.x)
             self.com_y_buffer.append(com_mark.pose.position.y)
 
             time_difference = (current_time - self.prev_t).to_sec()
 
-            # window length should be odd, greater then poly order and greater or equal to to the buffer size
+            # window length should be odd and greater than or equal to poly order
             if len(self.com_x_buffer) <= self.polyorder:
                 return self.marker
 
-            window_length = min(self.buffer_size, len(self.com_x_buffer))
-            window_length = window_length if window_length % 2 else window_length - 1
+            window_length = len(self.com_x_buffer)
+            window_length = window_length - 1 + window_length % 2
             x_dot = savgol_filter(self.com_x_buffer, window_length=window_length, polyorder=self.polyorder, deriv=1,
                                   delta=time_difference, mode='interp')[-1]
             y_dot = savgol_filter(self.com_y_buffer, window_length=window_length, polyorder=self.polyorder, deriv=1,
@@ -66,11 +65,8 @@ class CPCalculator(object):
 
                 x_cp = trans.transform.translation.x + x_dot * multiplier
                 y_cp = trans.transform.translation.y + y_dot * multiplier
-
                 self.update_marker(x_cp, y_cp)
-
                 self.prev_t = current_time
-
             except tf2_ros.TransformException as e:
                 rospy.logdebug('Error in trying to lookup transform for capture point: {error}'.format(error=e))
 
--- a/march_data_collector/src/march_data_collector/cp_calculator.py
+++ b/march_data_collector/src/march_data_collector/cp_calculator.py
@@ -50,10 +50,10 @@ class CPCalculator(object):
 
             window_length = min(self.buffer_size, len(self.com_x_buffer))
             window_length = window_length if window_length % 2 else window_length - 1
-            x_dot = savgol_filter(self.com_x_buffer, window_length=window_length, polyorder=4, deriv=1, delta=time_difference,
-                                  mode='interp')[-1]
-            y_dot = savgol_filter(self.com_y_buffer, window_length=window_length, polyorder=4, deriv=1, delta=time_difference,
-                                  mode='interp')[-1]
+            x_dot = savgol_filter(self.com_x_buffer, window_length=window_length, polyorder=self.polyorder, deriv=1,
+                                  delta=time_difference, mode='interp')[-1]
+            y_dot = savgol_filter(self.com_y_buffer, window_length=window_length, polyorder=self.polyorder, deriv=1,
+                                  delta=time_difference, mode='interp')[-1]
 
             try:
                 trans = self.tf_buffer.lookup_transform('world', self.foot_link, rospy.Time())
--- a/march_data_collector/src/march_data_collector/cp_calculator.py
+++ b/march_data_collector/src/march_data_collector/cp_calculator.py
@@ -17,7 +17,6 @@ class CPCalculator(object):
         self.prev_t = rospy.Time.now()
 
         self.marker = Marker()
-
         self.marker.header.frame_id = 'world'
         self.marker.type = self.marker.SPHERE
         self.marker.action = self.marker.ADD
@@ -31,6 +30,7 @@ class CPCalculator(object):
         self.g = 9.81  # gravity constant
         self.buffer_size = 25  # number of com's to use in estimation of velocity
         self.polyorder = 4  # order of the polynomial in Savitzky-Golay
+        self.time_difference = 0.04  # this is time interval between updates of the TF frames.
 
         self.com_x_buffer = RingBuffer(capacity=self.buffer_size, dtype=float64)
         self.com_y_buffer = RingBuffer(capacity=self.buffer_size, dtype=float64)
@@ -41,8 +41,6 @@ class CPCalculator(object):
             self.com_x_buffer.append(com_mark.pose.position.x)
             self.com_y_buffer.append(com_mark.pose.position.y)
 
-            time_difference = (current_time - self.prev_t).to_sec()
-
             # window length should be odd and greater than or equal to poly order
             if len(self.com_x_buffer) <= self.polyorder:
                 return self.marker
@@ -50,9 +48,9 @@ class CPCalculator(object):
             window_length = len(self.com_x_buffer)
             window_length = window_length - 1 + window_length % 2
             x_dot = savgol_filter(self.com_x_buffer, window_length=window_length, polyorder=self.polyorder, deriv=1,
-                                  delta=time_difference, mode='interp')[-1]
+                                  delta=self.time_difference, mode='interp')[-1]
             y_dot = savgol_filter(self.com_y_buffer, window_length=window_length, polyorder=self.polyorder, deriv=1,
-                                  delta=time_difference, mode='interp')[-1]
+                                  delta=self.time_difference, mode='interp')[-1]
 
             try:
                 trans = self.tf_buffer.lookup_transform('world', self.foot_link, rospy.Time())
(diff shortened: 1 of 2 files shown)


march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -1,6 +1,9 @@
 from math import sqrt
 
+from numpy import float64
```
Olavhaasie, line 3: You should add `<exec_depend>python-numpy</exec_depend>` to package.xml

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -1,6 +1,9 @@
...
+from numpy_ringbuffer import RingBuffer
 import rospy
+from scipy.signal import savgol_filter
```
Olavhaasie, line 6: You should add <exec_depend>python-scipy</exec_depend> to package.xml

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -29,13 +29,31 @@ def __init__(self, tf_buffer, foot_link):
         self.marker.scale.z = 0.03
 
         self.g = 9.81  # gravity constant
+        self.buffer_size = 25  # number of com's to use in estimation of velocity
+        self.polyorder = 4  # order of the polynomial in Saviztky-Golay
```
Olavhaasie, line 33:
```suggestion
        self.polyorder = 4  # order of the polynomial in Savitzky-Golay
```

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -29,13 +29,31 @@ def __init__(self, tf_buffer, foot_link):
...
         if current_time is not self.prev_t:
-            x_dot = (com_mark.pose.position.x - self.prev_x) / time_difference
-            y_dot = (com_mark.pose.position.y - self.prev_y) / time_difference
+
+            self.com_x_buffer.append(com_mark.pose.position.x)
+            self.com_y_buffer.append(com_mark.pose.position.y)
+
+            time_difference = (current_time - self.prev_t).to_sec()
+
+            # window length should be odd, greater then poly order and greater or equal to to the buffer size
```
Olavhaasie, line 47: Is the buffer size equal to the window length? Otherwise you could just this say, right?
```suggestion
            # window length should be odd and greater than or equal to the poly order
```

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -29,13 +29,31 @@ def __init__(self, tf_buffer, foot_link):
...
+            if len(self.com_x_buffer) <= self.polyorder:
+                return self.marker
+
+            window_length = min(self.buffer_size, len(self.com_x_buffer))
```
Olavhaasie, line 51: Isn't the definition of a ringbuffer that it cannot be greater than the given buffer size? So I think you can just use
```suggestion
            window_length = len(self.com_x_buffer)
```

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -29,13 +29,31 @@ def __init__(self, tf_buffer, foot_link):
...
+            window_length = window_length if window_length % 2 else window_length - 1
```
Olavhaasie, line 52: I find this easier to read instead of using the implicit 0 and 1 as truth values, see what you like I am giving you this power :battery: 
```suggestion
            window_length = window_length if window_length % 2 == 1 else window_length - 1
```

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -29,13 +29,31 @@ def __init__(self, tf_buffer, foot_link):
...
-            x_dot = (com_mark.pose.position.x - self.prev_x) / time_difference
-            y_dot = (com_mark.pose.position.y - self.prev_y) / time_difference
+
+            self.com_x_buffer.append(com_mark.pose.position.x)
+            self.com_y_buffer.append(com_mark.pose.position.y)
+
+            time_difference = (current_time - self.prev_t).to_sec()
+
+            # window length should be odd, greater then poly order and greater or equal to to the buffer size
+            if len(self.com_x_buffer) <= self.polyorder:
```
Olavhaasie, line 48: This is what you mean with the above comment right?
```suggestion
            if len(self.com_x_buffer) < self.polyorder:
```

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -29,13 +29,31 @@ def __init__(self, tf_buffer, foot_link):
...
+                return self.marker
+
+            window_length = min(self.buffer_size, len(self.com_x_buffer))
+            window_length = window_length if window_length % 2 else window_length - 1
+            x_dot = savgol_filter(self.com_x_buffer, window_length=window_length, polyorder=4, deriv=1, delta=time_difference,
```
Olavhaasie, line 53: You created a nice variable for this, but you don't use it :upside_down_face: 
```suggestion
            x_dot = savgol_filter(self.com_x_buffer, window_length=window_length, polyorder=self.polyorder, deriv=1, delta=time_difference,
```

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -29,13 +29,31 @@ def __init__(self, tf_buffer, foot_link):
...
+                                  mode='interp')[-1]
+            y_dot = savgol_filter(self.com_y_buffer, window_length=window_length, polyorder=4, deriv=1, delta=time_difference,
```
Olavhaasie, line 55:
```suggestion
            y_dot = savgol_filter(self.com_y_buffer, window_length=window_length, polyorder=self.polyorder, deriv=1, delta=time_difference,
```

march_data_collector/src/march_data_collector/cp_calculator.py
Roelemans: Or, if you want to make it less readable but more mathematical ;) 
```suggestion
            window_length = window_length - 1 + window_length % 2
```

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -29,13 +29,31 @@ def __init__(self, tf_buffer, foot_link):
...
+        self.com_x_buffer = RingBuffer(capacity=self.buffer_size, dtype=float64)
+        self.com_y_buffer = RingBuffer(capacity=self.buffer_size, dtype=float64)
 
     def calculate_cp(self, com_mark):
         current_time = com_mark.header.stamp
-        time_difference = (current_time - self.prev_t).to_sec()
         if current_time is not self.prev_t:
-            x_dot = (com_mark.pose.position.x - self.prev_x) / time_difference
-            y_dot = (com_mark.pose.position.y - self.prev_y) / time_difference
+
```
Roelemans, line 41:
```suggestion
```

march_data_collector/src/march_data_collector/cp_calculator.py
Roelemans, line 53: The time_difference is the time difference of the last measurement, but the filter will use it for all measurements before too right? Is it possible to pass a list similar to `com_x_buffer` for the time difference?

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -51,9 +69,8 @@ def calculate_cp(self, com_mark):
```
Roelemans, line 69:
```suggestion
```

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -51,9 +69,8 @@ def calculate_cp(self, com_mark):
 
                 self.update_marker(x_cp, y_cp)
 
-                self.prev_x = com_mark.pose.position.x
-                self.prev_y = com_mark.pose.position.y
                 self.prev_t = current_time
+
```
Roelemans, line 73: So many empty lines here :sweat_smile: 
```suggestion
```

march_data_collector/src/march_data_collector/cp_calculator.py
RutgerVanBeek: It makes no sense for us to have a buffer larger then the window length, so you are right.

march_data_collector/src/march_data_collector/cp_calculator.py
RutgerVanBeek: It should be greater then polynomial order, so when it is smaller or equal we escape.

march_data_collector/src/march_data_collector/cp_calculator.py
RutgerVanBeek: This was in a commit I forgot to push .....

RutgerVanBeek: I looked into this, but that is not possible. I have found the time difference to be quite stable. Furthermore it is not that fast as the ethercat cycle so small errors, won't have that much effect.

RutgerVanBeek:
> One other thing: I got very good experience with not smoothing velocity data. Did you consider trying to directly extract velocities from Gazebo? That should be mostly noiseless velocity data.

The algorithm is designed to not really have a time delay. You fit a polynomial which you evaluate at your current time. So I think that is much better than a moving average or exponential filter. I couldn't find a way to directly get the velocity of the center of mass of a model in Gazebo. You can get the velocities of the joints to calculate the change in center of mass, but then you miss the movement with respect to the world. So may be you can elaborate on what you mean?

march_data_collector/src/march_data_collector/cp_calculator.py
Roelemans: Won't it be better to keep it constant then? That way an outlier will have a smaller impact

march_data_collector/src/march_data_collector/cp_calculator.py
Olavhaasie: Then you should change to comment to remove _or equal to_

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -29,13 +29,30 @@ def __init__(self, tf_buffer, foot_link):
...
-        time_difference = (current_time - self.prev_t).to_sec()
         if current_time is not self.prev_t:
-            x_dot = (com_mark.pose.position.x - self.prev_x) / time_difference
-            y_dot = (com_mark.pose.position.y - self.prev_y) / time_difference
+            self.com_x_buffer.append(com_mark.pose.position.x)
+            self.com_y_buffer.append(com_mark.pose.position.y)
+
+            time_difference = (current_time - self.prev_t).to_sec()
+
+            # window length should be odd and greater than or equal to poly order
```
Olavhaasie, line 46:
```suggestion
            # window length should be odd and greater than poly order
```

RutgerVanBeek: Is this related to the comment on one of your other comments?

The window length should be strictly bigger than the poly order for mathematical reasons. So in the if-statement below we check whether it is smaller or equal and in that case we escape.

RutgerVanBeek:
@Roelemans and I had a discussion about the amount of noise and where it came from and whether it could be solved with making a plugin. 

Original situation:
Screenshot from 2020-06-24 11-33-15

Situation after fixing the frequency:
Screenshot from 2020-06-24 11-36-22

Situation with filter:
Screenshot from 2020-06-24 11-41-50

We see that 90% of the noise reduction was done by fixing the update frequency. I expect a lot more noise in real life, but we can also decide to add the filter only then.

Olavhaasie:
@RutgerVanBeek @Roelemans Is everything resolved now?

march_data_collector/src/march_data_collector/cp_calculator.py
RutgerVanBeek: I will still make it dependent on the publish rate of the joint state publisher, but keep it constant per run. 

> @RutgerVanBeek @Roelemans Is everything resolved now?

I am just pushing a commit on the time difference used.

RutgerVanBeek:
I have converted this into a draft, because the necessity and correctness of the filter is open for debate. We will keep discussing this and will also discuss this in the balance meeting.

PR #545, opened 2020-06-25
Feature/blc 41 Change structure to plan for both legs 

JorisWeeda:
Closes BLC-41

## Description
The first intention with this branch was to add the ending points of the non capture point move group, which uses the default subgait, as the start state to the capture point motion plan. This would increase the balance situation because the ending points would be taken into account when planning.

During this issue I've managed to plan for both legs using an `all_legs` move group which has the two legs as subgroups. If a left/right swing is requested the capture point leg will still plan as usual and the non capture point leg will get the default walk gait swing ending points. Setting targets for both the `left_leg/right_leg` move group legs the `all_legs` can create a trajectory for both legs. This would remove the need to set the start state and is in general way nicer because no merging of subgaits has to happen. :fire: 

This may sound a bit confusing so if any extra information is required don't hesitate to give me a call :calling: 

## Changes
* Removed merging of subgaits.
* Plan for both legs for the left and right swing


JorisWeeda:
> You should make a draft out of this

dick

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
```diff
@@ -20,19 +21,21 @@ def __init__(self, gait_name='gait_balanced_walk'):
         self._is_balance_used = rospy.get_param('/balance', False)
 
         self._move_group = None
+
         if self._is_balance_used:
             moveit_commander.roscpp_initialize(sys.argv)
             moveit_commander.RobotCommander()
             moveit_commander.PlanningSceneInterface()
```
RutgerVanBeek, line 28: Let's also add the ground plane here, we wil need it anyway.

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
```diff
@@ -73,10 +76,29 @@ def calculate_trajectory(self, leg_name):
...
-        return trajectory_plan.joint_trajectory
+    def calculate_normal_trajectory(self, leg_name, default_subgait):
+        """Calculate the pose of the non-capture-point leg and set this as the pose for this leg.
+
+        :param leg_name: The name of the move group which does not use capture point
+        :param default_subgait: The default subgait which is normally used
+        """
+        side_prefix = 'right' if 'right' in leg_name else 'left'
+
+        for joint in reversed(default_subgait.joints):
```
RutgerVanBeek, line 90: Bit of a 'beunfix' for editing a list while iterating over it.

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
```diff
@@ -108,75 +130,57 @@ def create_subgait_of_trajectory(normal_subgait, joint_trajectory):
...
+        targets = \
+            self._move_group['left_leg'].get_joint_value_target() + \
+            self._move_group['right_leg'].get_joint_value_target()
+
+        balance_subgait = self._move_group['all_legs'].plan(targets)
+        balance_trajectory = balance_subgait.joint_trajectory
+
+        if not balance_trajectory:
+            rospy.logwarn('No valid balance trajectory for {ln} received from capture point topic, '
+                          'returning default subgait'.format(ln=capture_point_leg_name))
```
RutgerVanBeek, line 161:
```suggestion
            rospy.logwarn('No valid balance trajectory for {ln} received from capture point leg, '
                          'returning default subgait'.format(ln=capture_point_leg_name))
```

Something like that, what does capture point topic might just be one of the reasons.

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
JorisWeeda: sad, I thought this was a compact and pretty good solution. Otherwise I have to create extra variables. Would your advice be to do something like this:
```
non_capture_point_joints = []
for joint in default_subgait.joints:
   if side_prefix in joint.name:
      non_capture_point_joints.append(joint)
```

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
JorisWeeda: I've tried adding the ground plane but the tests keep failing for some reason. To me it is a mystery so I am forced to leave the ground plane out of the balance gait for now

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
RutgerVanBeek: I think if someone changes the code in the future that it will soon do not what one would expect. I thought it was a bad habit to change something you are iterating over. Your suggestion is better I think.

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
RutgerVanBeek: Hmm I will look into that

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
JorisWeeda: suggestion is committed :+1:

JorisWeeda:
> When you launch without balance I get an error that the argument is not defined, did we define it in the simulation.launch?
> 
> I also got this error for each joint when MoveIt has planned a trajectory:
> "Could not interpolate setpoint at time 1.2005"

I to have noticed this error, not so sure why but will certainly look into it!

march_moveit/config/march.srdf
```diff
@@ -33,9 +33,13 @@
...
     </group>
+    <group name="all_legs">
+        <group name="left_leg" />
+        <group name="right_leg" />
+    </group>
     <!--END EFFECTOR: Purpose: Represent information about an end effector.-->
-    <end_effector name="right_foot" parent_link="foot_right" group="right_leg" />
-    <end_effector name="left_foot" parent_link="foot_left" group="left_leg" />
+    <end_effector name="foot_left" parent_link="foot_right" group="right_leg" />
+    <end_effector name="foot_left" parent_link="foot_left" group="left_leg" />
```
Roelemans, line 42: :thinking: IVI has two left ~hands~ feet

JorisWeeda: :ostrich:

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
JorisWeeda: maybe we forgot to import the right libraries in the cmake and package

PR #550, opened 2020-07-09
Update ankle tuning

Roelemans:
## Description

We get a lot of soft limit errors in de ankles. Today we tested some new tunings, but unfortunately made a mistake which meant the new tunings were not tested at all. In the end we only ran a test with p=300 and d=20. That went quite well, but p can probably be a lot higher still.


Ciliaclaij:
Are the values obtained with groundgaiting? Because then also the values in the groundgait file need to be adjusted

Deakensbubbles:
The values were indeed obtained through groundgaiting. I would also like to see these changes made in the groundgaiting file.

PR #551, opened 2020-07-13
Feature: divide subgaits

JorisWeeda:
Closes BLC-44

## Description
In order to have more control over the motion planning during the balance walk the subgait is divided into two sections, part one and part two. Part one is developed by the movement department and the second part is created using capture point in combination with the moveit motion planner. This structure should allow to generally define the path that the legs undertake during the balanced walk gait

## Changes
* Divided subgait into two parts
* Second part uses capture point motion planning
* Created new state machine for the divided subgaits.
* Added the balance gait files to the testing folder
* Updated all the tests with the newly added files

## Note
Don't forget to squash before merging (a lot of commits rip :blossom:)


march_state_machine/src/march_state_machine/state_machines/balance_walk_state_machine.py
```diff
@@ -0,0 +1,41 @@
...
+        """Initializes the walking pattern gait.
+
+        The gait will be initialized with the following subgaits:
+        * right_open
+        * right_swing
+        * left_swing
+        * right_close
+        * left_close
+        Where right_swing and left_swing can be stopped and will transition
+        to either right_close or left_close.
```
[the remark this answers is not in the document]
JorisWeeda: indeed, also added a little comment about why the subgaits are divided

march_state_machine/src/march_state_machine/state_machines/balance_walk_state_machine.py
```diff
@@ -9,13 +9,17 @@ def __init__(self, gait_name, check_gait_content=False):
...
+        * right_swing_2
+        * left_swing_1
+        * left_swing_2
         * right_close
         * left_close
-        Where right_swing and left_swing can be stopped and will transition
-        to either right_close or left_close.
+
+        The balanced walking pattern is identical to a normal walk but with divided subgaits.
+        This is necessar because capture point is used for the second part of the subgait.
```
RutgerVanBeek, line 22:
```suggestion
        This is necessary because capture point is used for the second part of the subgait.
```

march_gait_selection/test/nosetests/test_gait_selection.py
```diff
@@ -38,7 +38,8 @@ def test_types_in_loaded_gaits(self):
     # scan directory tests
     def test_scan_directory_top_level_content(self):
         directory = self.gait_selection.scan_directory()
-        self.assertEqual(['walk_medium', 'walk_small', 'stairs_up', 'walk'], directory.keys())
+        directory_gaits = ['walk_medium', 'balance_walk', 'stairs_up', 'walk_small', 'walk']
+        self.assertTrue(all(gait in directory_gaits for gait in directory.keys()))
```
Roelemans, line 42: Why not use assert equal here like before? That will give a more insightful error message when the test fails.

JorisWeeda: because travis CI produces a different order in comparison to local testing

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
```diff
@@ -17,9 +17,9 @@ class BalanceGait(object):
 
     def __init__(self, gait_name='gait_balanced_walk', move_groups=None):
         self.gait_name = gait_name
+        self.move_group = move_groups
```
Olavhaasie, line 20: Why did you make this member no longer protected?

JorisWeeda: because it is used to check if the balance gait is initiated

PR #552, opened 2020-07-13
Feature: gait graph format

Olavhaasie:
Closes PM-501

## Description
Right after the new subgait format, now comes the new gait graph format. Created a new `SubgaitGraph` class to do all the parsing and validating. It uses a breadth first algorithm to walk the graph and validate all transitions :sparkles: This way the `Gait` class was able to shrink by a lot and it was easy to write a lot of tests. I have also thought of extensibility, new types of transitions can effortlessly be added to the graph. 

See project-march/gait-files#191

## Changes
* New `SubgaitGraph` class and `SubgaitGraphError` exception
* New (parameterized) tests for the new functionality
* Parsing of new format
* Changed old formats used in testing


march_shared_classes/src/march_shared_classes/gait/subgait_graph.py
```diff
@@ -95,3 +95,9 @@ def __iter__(self):
                      for from_subgait, transitions in self._graph.items()
                      for to_subgait in transitions.values()
                      if len({from_subgait, to_subgait} & {self.START, self.END}) == 0])
+
+    def __eq__(self, other):
+        return isinstance(other, SubgaitGraph) and self._graph == other._graph
+
+    def __ne__(self, other):
+        return not self == other
```
Roelemans, line 103:
```suggestion
        return not self.__eq__(other)
```
In https://github.com/project-march/march/blob/develop/march_shared_classes/src/march_shared_classes/gait/setpoint.py we do it like this so maybe keep that format for consistency?

PR #553, opened 2020-07-14
Feature/blc 47 predict falling formulas

Roelemans:
Closes BLC-47

## Description

Capture point changes as exo falls when it lifts one foot. Hence, one wants to step to the location where capture point will be ~1 second in the future. This PR introduces a numerical method that predicts how the CoM will fall as exo falls. It returns the position and speed of the CoM at t seconds in the future. This can then be used to calculate the capture point at that time. See https://www.overleaf.com/project/5f0467b1e6672b00015e27b8 (or the [pdf](https://github.com/project-march/march/files/4917392/Balance___inverted_pendulum_calculations.pdf) )for the mathematical details (you can skip chapter 3.1 as that is about a Taylor expansion that did not converge).

## Changes

* Added a method for CoM path prediction


march_gait_selection/src/march_gait_selection/dynamic_gaits/inverted_pendulum.py
```diff
@@ -0,0 +1,38 @@
+import math
+
+
+class InvertedPendulum(object):
+    g = 9.81
```
Olavhaasie, line 5: Class constants are usually all upper case
```suggestion
    G = 9.81
```

march_gait_selection/src/march_gait_selection/dynamic_gaits/inverted_pendulum.py
```diff
@@ -0,0 +1,38 @@
...
+
+    @classmethod
+    def numeric_solve_to_t(cls, x, y, z, vx, vy, t, dt=0.0001):
+        time = 0
+        while time < t:
+            x, y, z, vx, vy = cls.step_numeric_solve(x, y, z, vx, vy, dt)
+            time += dt
+        if abs(z) < 0.000001:
```
Olavhaasie, line 13: You could turn this number into a class constant `EPSILON`. Makes it easier to change and use in the future.

march_gait_selection/src/march_gait_selection/dynamic_gaits/inverted_pendulum.py
Roelemans: Whoops

march_gait_selection/src/march_gait_selection/dynamic_gaits/inverted_pendulum.py
Roelemans: Ah good point. Thanks for the feedback!

march_gait_selection/src/march_gait_selection/dynamic_gaits/inverted_pendulum.py
```diff
@@ -0,0 +1,39 @@
...
+    G = 9.81
+    EPSILON = 0.000001
+
+    @classmethod
+    def numeric_solve_to_t(cls, x, y, z, vx, vy, t, dt=0.01):
+        time = 0
+        while time < t:
+            x, y, z, vx, vy = cls.step_numeric_solve(x, y, z, vx, vy, dt)
+            time += dt
+        if abs(z) < EPSILON:
```
Olavhaasie, line 14: Shouldn't this be?
```suggestion
        if abs(z) < cls.EPSILON:
```

Roelemans: I am deeply ashamed of my incompetence

RutgerVanBeek:
I have no comments on the implementation, but it looks like the backwards Euler scheme you use is not the same as you derive in the pdf. You use some different notation, so may be you can update either one or explain it to me.

PR #555, opened 2020-07-16
decrease effort

RutgerVanBeek:
## Description

It turns out that the ability of the iMC to follow our effort command decreases a lot when we command high efforts. This means that when we do max effort for a long time the actual effort can go anywhere. As for most gaits 10000 effort is enough, we can decrease the limit a bit, to reduce the chance of a short-circuit error. I have kept the limit for the linear joints the same. In theory they could encounter the same problem. In practice they never reach that amount of effort. 
## Changes

Decrease the effort limit for rotary joints.

### DIFF
--- a/march_description/urdf/march4.xacro
+++ b/march_description/urdf/march4.xacro
@@ -19,7 +19,9 @@
         </gazebo>
     </xacro:if>
 
-    <xacro:arg name="max_effort_rotary" default="20475.0" />  <!-- = 25 A -->
+    <!-- The max effort of the rotary joints is lower then the value corresponding to 25A, because it has been shown
+    that the chances of a short-circuit when applying the theoretical max effort for a long time are significant. -->
+    <xacro:arg name="max_effort_rotary" default="18000.0" />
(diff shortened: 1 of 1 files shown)


march_description/urdf/march4.xacro
```diff
@@ -19,7 +19,9 @@
         </gazebo>
     </xacro:if>
 
-    <xacro:arg name="max_effort_rotary" default="20475.0" />  <!-- = 25 A -->
+    <!-- The max effort of the rotary joints is lower then the value corresponding to 25A, because it has been shown
+    that the chances of a short-circuit when applying the theoretical max effort for a long time are significant. -->
```
Roelemans, line 23:
```suggestion
    <!-- The max effort of the rotary joints is lower then the value corresponding to 25A, because short-circuit errors
    have occured while applying the theoretical max effort for a long time. -->
```
¨it has been shown" makes it seem like it is unavoidable, but it might be caused by faulty hardware or may very well be fixed by retuning current controller.

RutgerVanBeek: fair enough

PR #556, opened 2020-07-17
Created new state machine for a fixed number of steps on the ramp down

marnixbrands:
## Description

Last training, Sjaan indicated that it would be nice if there is a fixed number of steps preprogrammed on the ramp, so she doesn't have to trigger the IPD for a closing step. So I added a new state machine named `slope_down_fixed_sm` in which the first four steps are fixed (right_open, left_swing, right_swing and left_close). After these four steps, you're in the state "STANDING SLOPE DOWN", so that you can execute the last step off the ramp or an additional single step on the ramp if necessary. I set it on four fixed steps for now, but since we don't now yet how many steps Sjaan needs on the ramp, it can be extended to six steps (two extra swings). 
## Changes

Added a new state machine named `slope_down_fixed_sm`

### DIFF
--- a/march_state_machine/src/march_state_machine/healthy_sm.py
+++ b/march_state_machine/src/march_state_machine/healthy_sm.py
@@ -4,8 +4,8 @@ from std_srvs.srv import Empty, EmptyRequest
 
 from march_shared_resources.srv import CurrentState, PossibleGaits
 
-from .gaits import slope_down_sm
 from .gaits import slope_down_fixed_sm
+from .gaits import slope_down_sm
 from .gaits import tilted_path_left_flexed_knee_step_sm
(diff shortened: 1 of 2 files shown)


march_state_machine/src/march_state_machine/healthy_sm.py
```diff
@@ -5,6 +5,7 @@
 from march_shared_resources.srv import CurrentState, PossibleGaits
 
 from .gaits import slope_down_sm
+from .gaits import slope_down_fixed_sm
```
Olavhaasie, line 8:
```suggestion
from .gaits import slope_down_fixed_sm
from .gaits import slope_down_sm
```

PR #557, opened 2020-07-22
Feature: PM-511 parametric gait versions

RutgerVanBeek:
Closes PM-511

## Description

With https://github.com/project-march/monitor/pull/63 parametric gait versions are created. These versions are of the form: `{parameter}({base version})_({other version})`. Here the Subgait and Gait classes are changed to handle this. The gait selection is also updated. I have moved all the splitting of versions to the Subgait class, so other classes can just use the new versions strings as if nothing changed. This meant that something functionality had to move to the Subgait class.
## Changes

* Changed the Subgait class to handel parametric gait versions
* Added a method to get a Subgait from name and versions instead of from a file path.

## Testing
Use the branch in the monitor repo to get parametric gait versions. You can also put a parametric gait version in the `default.yaml`.

### DIFF
--- a/march_shared_classes/src/march_shared_classes/gait/subgait.py
+++ b/march_shared_classes/src/march_shared_classes/gait/subgait.py
@@ -12,7 +12,7 @@ from .joint_trajectory import JointTrajectory
 from .limits import Limits
 from .setpoint import Setpoint
 
-PARAMETRIC_GAITS_CHARACTER = '_'
+PARAMETRIC_GAITS_PREFIX = '_pg_'
 
 
 class Subgait(object):
@@ -73,7 +73,7 @@ class Subgait(object):
         :param args:
         :return: A populated Subgait object.
         """
-        if version.startswith(PARAMETRIC_GAITS_CHARACTER):
+        if version.startswith(PARAMETRIC_GAITS_PREFIX):
             base_version, other_version, parameter = Subgait.unpack_parametric_version(version)
             base_path = os.path.join(gait_dir, gait_name, subgait_name, base_version + '.subgait')
             other_path = os.path.join(gait_dir, gait_name, subgait_name, other_version + '.subgait')
@@ -274,7 +274,7 @@ class Subgait(object):
 
         duration = base_subgait.duration * parameter + (1 - parameter) * other_subgait.duration
         gait_type = base_subgait.gait_type if parameter <= 0.5 else other_subgait.gait_type
-        version = '{0}{1}_({2})_({3})'.format(PARAMETRIC_GAITS_CHARACTER, parameter, base_subgait.version,
+        version = '{0}{1}_({2})_({3})'.format(PARAMETRIC_GAITS_PREFIX, parameter, base_subgait.version,
                                               other_subgait.version)
         return Subgait(joints, duration, gait_type, base_subgait.gait_name, base_subgait.subgait_name, version,
                        description)
@@ -334,7 +334,7 @@ class Subgait(object):
     @staticmethod
     def validate_version(gait_path, subgait_name, version):
         """Check whether a gait exists for the gait."""
-        if version[0] == PARAMETRIC_GAITS_CHARACTER:
+        if version.startswith(PARAMETRIC_GAITS_PREFIX):
             base_version, other_version, _ = Subgait.unpack_parametric_version(version)
             base_subgait_path = os.path.join(gait_path, subgait_name, base_version + '.subgait')
             other_subgait_path = os.path.join(gait_path, subgait_name, other_version + '.subgait')
@@ -352,7 +352,7 @@ class Subgait(object):
     @staticmethod
     def unpack_parametric_version(version):
         """Unpack a version to base version, other version and parameter."""
-        parameter_str = re.search(r'{0}[0-9.]*_'.format(PARAMETRIC_GAITS_CHARACTER), version).group(0)
+        parameter_str = re.search(r'{0}[0-9.]*_'.format(PARAMETRIC_GAITS_PREFIX), version).group(0)
         parameter = float(parameter_str[1:-1])
         versions = re.findall(r'\([^\)]*\)', version)
         base_version = versions[0][1:-1]
--- a/march_shared_classes/src/march_shared_classes/gait/subgait.py
+++ b/march_shared_classes/src/march_shared_classes/gait/subgait.py
@@ -352,8 +352,8 @@ class Subgait(object):
     @staticmethod
     def unpack_parametric_version(version):
         """Unpack a version to base version, other version and parameter."""
-        parameter_str = re.search(r'{0}[0-9.]*_'.format(PARAMETRIC_GAITS_PREFIX), version).group(0)
-        parameter = float(parameter_str[1:-1])
+        parameter_str = re.search(r'^{0}(\d+\.\d+)_'.format(PARAMETRIC_GAITS_PREFIX), version).group(1)
+        parameter = float(parameter_str)
         versions = re.findall(r'\([^\)]*\)', version)
         base_version = versions[0][1:-1]
         other_version = versions[1][1:-1]
--- a/march_shared_classes/src/march_shared_classes/gait/subgait.py
+++ b/march_shared_classes/src/march_shared_classes/gait/subgait.py
@@ -12,7 +12,7 @@ from .joint_trajectory import JointTrajectory
 from .limits import Limits
 from .setpoint import Setpoint
 
-PARAMETRIC_GATIS_CHARACTER = '_'
+PARAMETRIC_GAITS_CHARACTER = '_'
 
 
 class Subgait(object):
@@ -45,11 +45,8 @@ class Subgait(object):
         :returns
             A populated Subgait object
         """
-        if file_name is None or file_name == '':
+        if file_name is None or not os.path.isfile(file_name):
             raise FileNotFoundError(file_path=file_name)
-        if not os.path.isfile(file_name):
-            raise FileNotFoundError(file_path=file_name)
-
         try:
             gait_name = file_name.split('/')[-3]
             subgait_name = file_name.split('/')[-2]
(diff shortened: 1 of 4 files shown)


march_shared_classes/src/march_shared_classes/gait/subgait.py
```diff
@@ -40,7 +46,9 @@ def from_file(cls, robot, file_name, *args):
             A populated Subgait object
         """
         if file_name is None or file_name == '':
-            return None
+            raise FileNotFoundError(file_path=file_name)
+        if not os.path.isfile(file_name):
+            raise FileNotFoundError(file_path=file_name)
```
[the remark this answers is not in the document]
RutgerVanBeek: I got an error with `os.path.isfile(None), so I thought you had to check that first.`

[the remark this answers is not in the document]
Olavhaasie: Not if you check `file_name is not None` before that :wink:

march_shared_classes/src/march_shared_classes/gait/subgait.py
```diff
@@ -1,13 +1,19 @@
...
 import yaml
 
 from march_shared_classes.exceptions.gait_exceptions import NonValidGaitContent, SubgaitInterpolationError
+from march_shared_classes.exceptions.general_exceptions import FileNotFoundError
 
 from .joint_trajectory import JointTrajectory
 from .limits import Limits
 from .setpoint import Setpoint
 
+PARAMETRIC_GATIS_CHARACTER = '_'
```
JorisWeeda, line 15:
```suggestion
PARAMETRIC_GAITS_CHARACTER = '_'
```

march_shared_classes/src/march_shared_classes/gait/subgait.py
```diff
@@ -301,3 +330,31 @@ def __getitem__(self, index):
     def __len__(self):
         return len(self.joints)
     # endregion
+
+    @staticmethod
+    def validate_version(gait_path, subgait_name, version):
+        """Check whether a gait exists for the gait."""
+        if version[0] == PARAMETRIC_GATIS_CHARACTER:
+            base_version, other_version, _ = unpack_parametric_version(version)
```
JorisWeeda, line 338: should it not be?
```suggestion
            base_version, other_version, _ = Subgait.unpack_parametric_version(version)
```

march_shared_classes/src/march_shared_classes/gait/subgait.py
```diff
@@ -56,6 +64,27 @@ def from_file(cls, robot, file_name, *args):
...
+
+        :param robot: The robot corresponding to the given subgait file
+        :param gait_dir: The directory with all the gaits
+        :param gait_name: The name of the corresponding gait
+        :param subgait_name: The name of the subgait to load
+        :param version: The version to use, this can be parametric
+        :param args:
+        :return: A populated Subgait object.
+        """
+        if version[0] == PARAMETRIC_GATIS_CHARACTER:
```
JorisWeeda, line 79: so the subgait is a parametric subgait if the prefix of the subgait name is `_` ?

march_shared_classes/src/march_shared_classes/gait/subgait.py
```diff
@@ -301,3 +330,31 @@ def __getitem__(self, index):
...
+                    return False
+        else:
+            subgait_path = os.path.join(gait_path, subgait_name, version + '.subgait')
+            if not os.path.isfile(subgait_path):
+                rospy.logwarn('{sp} does not exist'.format(sp=subgait_path))
+                return False
+        return True
+
+
+def unpack_parametric_version(version):
```
JorisWeeda, line 353: this could also be private or a static private class method, maybe nice to privatize it because it is rather specific to the subgait class and is not used outside of this file

march_shared_classes/src/march_shared_classes/gait/subgait.py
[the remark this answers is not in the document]
RutgerVanBeek: Does it work like that when you put them on the same line?

march_shared_classes/src/march_shared_classes/gait/subgait.py
RutgerVanBeek: I know, but I want to keep it a string as that makes live a lot easier. I can also disable the possibility in the gait generator to save a gait with version starting with underscore.

march_shared_classes/src/march_shared_classes/gait/subgait.py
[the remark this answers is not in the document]
Olavhaasie: Yes the python `and` will "short circuit". If the first expression evaluates to `False` then there is no reason in evaluating the second expression, since the result will always be `False`.

march_shared_classes/src/march_shared_classes/gait/subgait.py
JorisWeeda: u can make it a staticmethod for the subgait class and call it using `Subgait.unpack_parametric_version`. I use similar coding in the transition subgait. or just put `_` in front to make it private maybe

march_shared_classes/src/march_shared_classes/gait/subgait.py
JorisWeeda: hmm, I guess a proper prefix would save a lot of problems. Maybe extend the prefix to `_parametric_` ?

RutgerVanBeek: You already get a very long expression in the menu, I think a short prefix is better, but I will change it if you both f=think `_parametric_` is better.

march_shared_classes/src/march_shared_classes/gait/subgait.py
RutgerVanBeek: Your first suggestion is the prettiest I think.

march_shared_classes/src/march_shared_classes/gait/subgait.py
JorisWeeda: hmm, I guess both would be fine if we explicitly define it somewhere, maybe in the tutorials

JorisWeeda: or we do something like `_pg_` as a prefix and call the parameter
```
PARAMETRIC_GAIT_PREFIX =  '_pg_'
```

march_shared_classes/src/march_shared_classes/gait/subgait.py
JorisWeeda: nice

march_shared_classes/src/march_shared_classes/gait/subgait.py
```diff
@@ -301,3 +327,31 @@ def __getitem__(self, index):
     def __len__(self):
         return len(self.joints)
     # endregion
+
+    @staticmethod
+    def validate_version(gait_path, subgait_name, version):
+        """Check whether a gait exists for the gait."""
+        if version[0] == PARAMETRIC_GAITS_CHARACTER:
```
Olavhaasie, line 334:
```suggestion
        if version.starts_with(PARAMETRIC_GAITS_CHARACTER):
```

march_shared_classes/src/march_shared_classes/gait/subgait.py
```diff
@@ -1,13 +1,19 @@
...
 import yaml
 
 from march_shared_classes.exceptions.gait_exceptions import NonValidGaitContent, SubgaitInterpolationError
+from march_shared_classes.exceptions.general_exceptions import FileNotFoundError
 
 from .joint_trajectory import JointTrajectory
 from .limits import Limits
 from .setpoint import Setpoint
 
+PARAMETRIC_GAITS_CHARACTER = '_'
```
Olavhaasie, line 15: It might be nice to name it a prefix to not limit it to one character
```suggestion
PARAMETRIC_GAITS_PREFIX = '_'
```

march_shared_classes/src/march_shared_classes/gait/subgait.py
```diff
@@ -247,7 +273,10 @@ def interpolate_subgaits(cls, base_subgait, other_subgait, parameter):
             base_subgait.version, other_subgait.version, parameter)
 
         duration = base_subgait.duration * parameter + (1 - parameter) * other_subgait.duration
-        return Subgait(joints, duration, base_subgait.gait_name, base_subgait.subgait_name, 'interpolated subgait',
+        gait_type = base_subgait.gait_type if parameter <= 0.5 else other_subgait.gait_type
```
Olavhaasie, line 276: So if the parameter is 0.0 and the base_subgait is not used then it should the gait type of the base gait? It sounds like that should be the other way around
```suggestion
        gait_type = base_subgait.gait_type if parameter > 0.5 else other_subgait.gait_type
```

march_shared_classes/src/march_shared_classes/gait/subgait.py
```diff
@@ -301,3 +330,31 @@ def __getitem__(self, index):
...
+            if not os.path.isfile(subgait_path):
+                rospy.logwarn('{sp} does not exist'.format(sp=subgait_path))
+                return False
+        return True
+
+    @staticmethod
+    def unpack_parametric_version(version):
+        """Unpack a version to base version, other version and parameter."""
+        parameter_str = re.search(r'{0}[0-9.]*_'.format(PARAMETRIC_GAITS_PREFIX), version).group(0)
+        parameter = float(parameter_str[1:-1])
```
Olavhaasie, line 356: Because the prefix can now be multiple characters, it is nicer to match a group instead of matching the entire string and getting a substring starting from 1. Furthermore, the float regex could be empty. I replaced it with another regex that assumes you don't use negative parameters and doesn't allow floats like 0. and .0
```suggestion
        parameter_str = re.search(r'^{0}(\d+\.\d+)_'.format(PARAMETRIC_GAITS_PREFIX), version).group(1)
        parameter = float(parameter_str)
```

march_shared_classes/src/march_shared_classes/gait/subgait.py
RutgerVanBeek: parameter equaling zero means the base subgait, parameter equaling one means the other subgait. With `parameter==0` the base_subgait is returned.

march_shared_classes/src/march_shared_classes/gait/subgait.py
RutgerVanBeek: Thanks I don't know regex that well.

PR #560, opened 2020-07-23
round duration in joint trajectory

RutgerVanBeek:
Closes PM-532

## Description

The attributes of a setpoint are rounded up to four digits. However the duration of the joint trajectory was not. The result of this is that it could happend that the time at which we requested an interpolated setpoint was a bit after the duration. Normally it is not possible to do this, therefore an error was logged. This is fixed by also rounding the duration in the `JointTrajectory` class.

## Changes

Round the duration in the `JointTrajectory` class.

## Testing
The close versions of the walk gait that are default in airgait-v and training-v, would cause this error normally.

### DIFF
--- a/march_shared_classes/src/march_shared_classes/gait/joint_trajectory.py
+++ b/march_shared_classes/src/march_shared_classes/gait/joint_trajectory.py
@@ -57,7 +57,7 @@ class JointTrajectory(object):
                 if setpoint.time > new_duration:
                     self.setpoints.remove(setpoint)
 
-        self._duration = round(new_duration, self.self.setpoint_class.digits)
+        self._duration = round(new_duration, self.setpoint_class.digits)
         self.interpolate_setpoints()
 
     @property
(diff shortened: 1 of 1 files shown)


march_shared_classes/src/march_shared_classes/gait/joint_trajectory.py
```diff
@@ -10,12 +10,13 @@ class JointTrajectory(object):
...
 
     setpoint_class = Setpoint
+    digits = 4
 
     def __init__(self, name, limits, setpoints, duration, *args):
         self.name = name
         self.limits = limits
         self._setpoints = setpoints
-        self._duration = duration
+        self._duration = round(duration, self.digits)
```
JorisWeeda, line 19:
```suggestion
        self._duration = round(duration, Setpoint.digits)
```
This way we only have to manage the variable in the setpoint class

march_shared_classes/src/march_shared_classes/gait/joint_trajectory.py
```diff
@@ -10,12 +10,13 @@ class JointTrajectory(object):
     """Base class for joint trajectory of a gait."""
 
     setpoint_class = Setpoint
+    digits = 4
```
JorisWeeda, line 13: can't this be deleted?
```suggestion
```

RutgerVanBeek: of course

PR #561, opened 2020-07-24
Add arg for data collector

Olavhaasie:
## Description
Since the data collector is not a necessary node, I have added an argument that enables it.

## Changes
* add `data_collector` arg

### DIFF
--- a/march_launch/launch/march.launch
+++ b/march_launch/launch/march.launch
@@ -71,7 +71,7 @@
             <arg name="ping_safety_node" value="true"/>
         </include>
 
-        <include file="$(find march_data_collector)/launch/march_data_collector.launch" if="$(eval data_collector or esp)">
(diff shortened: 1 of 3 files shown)


march_launch/launch/march.launch
```diff
@@ -70,13 +71,13 @@
             <arg name="ping_safety_node" value="true"/>
         </include>
 
-        <include file="$(find march_data_collector)/launch/march_data_collector.launch">
+        <include file="$(find march_data_collector)/launch/march_data_collector.launch" if="$(eval data_collector or esp)">
```
JorisWeeda, line 74:
```suggestion
        <include file="$(find march_data_collector)/launch/march_data_collector.launch" if="$(eval data_collector or esp or balance)">
```
Data collector node is necessary to calculate capture point

PR #563, opened 2020-07-27
Feature/blc 48 capture point service call

JorisWeeda:
Closes BLC-48

## Description
Create a service setup to request the capture point position over a given duration. It uses the inverted pendulum model the calculate the position of the center of mass over a certain duration. This new CoM position is used to calculate the respected capture point pose of one of the move groups.

## Changes
* Create service side on the data collector
* Created a new message format for the capture point pose
* Requested capture point in the gait selection balance gait

## Testing
To test this use the `minimal` gait directory which is the default when launching the simulation.


march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -1,70 +1,114 @@
...
-
-                x_cp = trans.transform.translation.x + x_dot * multiplier
-                y_cp = trans.transform.translation.y + y_dot * multiplier
-
-                self.update_marker(x_cp, y_cp)
-
-                self.prev_x = com_mark.pose.position.x
-                self.prev_y = com_mark.pose.position.y
-                self.prev_t = current_time
+                world_transform = self._tf_buffer.lookup_transform('world', self._foot_link, rospy.Time())
```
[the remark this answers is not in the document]
Roelemans: Indeed this should simply be the new position of the center of mass (see suggestion below)

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -1,70 +1,114 @@
...
+                if new_center_of_mass.z <= 0:
+                    rospy.logdebug_throttle(1, 'Cannot calculate capture point; center of mass < 0')
+
+                capture_point_multiplier = sqrt(new_center_of_mass.z / self._gravity_constant)
+
+                x_cp = world_transform.transform.translation.x + new_center_of_mass.vx * capture_point_multiplier
+                y_cp = world_transform.transform.translation.y + new_center_of_mass.vy * capture_point_multiplier
```
Roelemans, line 92:
```suggestion
                x_cp = new_center_of_mass.x + new_center_of_mass.vx * capture_point_multiplier
                y_cp = new_center_of_mass.y + new_center_of_mass.vy * capture_point_multiplier
```

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -1,70 +1,114 @@
...
+    def _calculate_capture_point(self, duration):
+        """Calculate a future capture point pose using the inverted pendulum and center of mass.
+
+        :param duration:
+            the amount of seconds away from the current time the capture point should be calculated
+        """
+        rospy.logdebug('time diff: {td}'.format(td=str(self._delta_t)))
+        if self._delta_t == 0:
+            rospy.loginfo('Calculating future center of mass for capture point pose.')
+            new_center_of_mass = InvertedPendulum.numeric_solve_to_t(self.x, self.y, self.z, self.vx, self.vy, duration)
```
Roelemans, line 82:
```suggestion
            new_center_of_mass = InvertedPendulum.numeric_solve_to_t(self.x - world_transform.transform.translation.x, self.y - world_transform.transform.translation.y, self.z - world_transform.transform.translation.z, self.vx, self.vy, duration)
```
These coordinates shoulde be relative to the stable foot.

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -1,70 +1,114 @@
...
+                    rospy.logdebug_throttle(1, 'Cannot calculate capture point; center of mass < 0')
+
+                capture_point_multiplier = sqrt(new_center_of_mass.z / self._gravity_constant)
+
+                x_cp = world_transform.transform.translation.x + new_center_of_mass.vx * capture_point_multiplier
+                y_cp = world_transform.transform.translation.y + new_center_of_mass.vy * capture_point_multiplier
+
+                self._capture_point_marker.header.stamp = rospy.get_rostime()
+                self._capture_point_marker.pose.position.x = x_cp
+                self._capture_point_marker.pose.position.y = y_cp
```
Roelemans, line 96:
```suggestion
                self._capture_point_marker.pose.position.x = x_cp + world_transform.transform.translation.x
                self._capture_point_marker.pose.position.y = y_cp + world_transform.transform.translation.y
```
And transform them back to the world frame

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
```diff
@@ -158,6 +162,7 @@ def create_subgait_of_trajectory(self, joint_trajectory, subgait_name):
             balance_joints.append(JointTrajectory(joint_name, normal_joint.limits, setpoints, balance_duration))
 
         balance_subgait = BalanceGait.to_subgait(balance_joints, balance_duration, subgait_name=subgait_name)
+        self.export_to_file(balance_subgait, os.path.dirname(os.path.realpath(__file__)))
```
Olavhaasie, line 165: Why this writing to a file?

march_shared_resources/srv/CapturePointPose.srv
```diff
@@ -0,0 +1,10 @@
+# Service for setting subgait versions of a gait.
+
+string duration
```
Olavhaasie, line 3: Why is the duration a string? Isn't that quite confusing as to what format it should take?

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
Roelemans: So that we can have a look at the gait Moveit produced after the execution by opening it in the gait generator. Very useful for testing purposes.

march_shared_resources/srv/CapturePointPose.srv
Roelemans: Changed the duration and the xyz below to float64 :+1:

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -1,70 +1,152 @@
...
-    def update_marker(self, x_cp, y_cp):
-        self.marker.header.stamp = rospy.get_rostime()
-        self.marker.pose.position.x = x_cp
-        self.marker.pose.position.y = y_cp
-        self.marker.pose.position.z = 0
-
-        rospy.logdebug('capture point is at ' + str(self.marker.pose.position))
-
-        self.publisher.publish(self.marker)
+    def __init__(self, tf_buffer, swing_foot_link, static_foot_link):
```
RutgerVanBeek, line 14: For the entire class: all references to tf or different feet  can be removed.

march_data_collector/src/march_data_collector/data_collector_node.py
```diff
@@ -174,7 +174,10 @@ def main():
     tf_buffer = tf2_ros.Buffer()
     tf2_ros.TransformListener(tf_buffer)
     center_of_mass_calculator = CoMCalculator(robot, tf_buffer)
-    feet = ['foot_left', 'foot_right']
-    cp_calculators = [CPCalculator(tf_buffer, foot) for foot in feet]
+
+    # key is the swing foot and item is the static foot
+    feet = {'foot_left': 'foot_right', 'foot_right': 'foot_left'}
```
RutgerVanBeek, line 179: Aren't there better ways to do this?

march_data_collector/src/march_data_collector/data_collector_node.py
```diff
@@ -174,7 +174,10 @@ def main():
...
+    cp_calculators = [CPCalculator(tf_buffer, swing_foot, static_foot) for swing_foot, static_foot in feet.items()]
+
```
RutgerVanBeek, line 181: We only need one now right?

march_shared_resources/srv/CapturePointPose.srv
```diff
@@ -0,0 +1,10 @@
+# Service for getting the position of the center of mass after falling for time duration
+
+float64 duration
+---
+bool success
+string message
```
RutgerVanBeek, line 6: may be a more descriptive name then message

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -1,70 +1,148 @@
...
+
+    def _calculate_capture_point(self, duration):
+        """Calculate a future capture point pose using the inverted pendulum and center of mass.
+
+        :param duration:
+            the amount of seconds away from the current time the capture point should be calculated
+        """
+        try:
+            world_transform = self._tf_buffer.lookup_transform('world', self._swing_foot_link, rospy.Time())
+
```
RutgerVanBeek, line 87: @Roelemans How does this work? Why do we still need the position relative to the swing foot link. Do we use the inverted pendulum such that we need the swing leg foot as base and then need the relative position of the com, as an absolute position in the inverted pendulum frame?

march_data_collector/src/march_data_collector/data_collector_node.py
RutgerVanBeek: @Roelemans ?

march_data_collector/src/march_data_collector/plot_pendulum.py
```diff
@@ -0,0 +1,100 @@
...
+# print("num deriv o2 = ", pendulum.numeric_derivative(2, x0, y0, z0, x1, y1, dt))
+# print("num deriv o3 = ", pendulum.numeric_derivative(3, x0, y0, z0, x1, y1, dt))
+# print("num deriv o4 = ", pendulum.numeric_derivative(4, x0, y0, z0, x1, y1, dt))
+# print("num deriv o5 = ", pendulum.numeric_derivative(5, x0, y0, z0, x1, y1, dt))
+# print("num deriv o6 = ", pendulum.numeric_derivative(6, x0, y0, z0, x1, y1, dt))
+# print("num deriv o7 = ", pendulum.numeric_derivative(7, x0, y0, z0, x1, y1, dt))
+
+plot = 1
+if plot:
+    num_t = [0]
```
RutgerVanBeek, line 40: @Roelemans do you think this plot functionality should be included in the PR? I will clean the code if you think it should be included.

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
```diff
@@ -210,3 +230,22 @@ def __getitem__(self, name):
             return self.construct_subgait('right_leg', name)
         else:
             return self.default_walk[name]
+
+    @staticmethod
+    def export_to_file(subgait, gait_directory):
+        if gait_directory is None or gait_directory == '':
+            return
```
RutgerVanBeek, line 237: @Roelemans if you think this is necessary I will move it to the subgait class otherwise I will remove this.

march_data_collector/src/march_data_collector/plot_pendulum.py
Roelemans: No that was accidentally committed; I removed it

march_data_collector/src/march_data_collector/cp_calculator.py
Roelemans: Ah no, this is a naming issue (git blame @JorisWeeda); what we call swing foot here is actually the static foot, and the other way around. We call the service "/march/cp_marker_{static_foot}" which we interpret as the the topic for the swing foot elsewhere in the code. I will fix this.

march_data_collector/src/march_data_collector/data_collector_node.py
Roelemans: Hmm in the current implementation we need two; because both need the static_foot_link for a different foot. However, you could implement it in such a way that you can make both services in a single instance of the class

march_data_collector/src/march_data_collector/data_collector_node.py
Roelemans: Probably, feel free to cook something up

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
Roelemans: I think this is a very useful diagnostic tool! Why move it upstream? The one in the gait_generator_controller.py is somewhat different and calls some methods from rqt :s

RutgerVanBeek: I think this method does not belong here, why not make a method write_to_file(self, path) in the subgait class

Roelemans:
I think it would be more useful to close this and open a new one since the description and the comment history are outdated

PR #567, opened 2020-07-31
[gait-selection-develop] Feature: error handling

Olavhaasie:
Closes PM-507

## Description
The new gait selection state machine now responds to the `/march/error` topic and stops the current gait when a non fatal error is published and shuts down when a fatal error is published.

## Changes
* Add error subscriber
* Add `stop` method

## Testing
Start a `roscore`. Open another shell and upload the URDF and joint names (necessary for safety node)
```
roslaunch march_launch upload_march_urdf.launch
rosrun march_simulation upload_joint_names
```
Now launch the gait selection, safety and possibly an input device. To see how it responds. You can checkout `remove-gait-prefix` in project-march/monitor to be able to issue gait commands.

### DIFF
--- a/march_gait_selection/src/march_gait_selection/gait_selection_node.py
+++ b/march_gait_selection/src/march_gait_selection/gait_selection_node.py
@@ -1,6 +1,7 @@
 import rospy
 from std_srvs.srv import Trigger
 
+from march_shared_resources.msg import Error
 from march_shared_resources.srv import (ContainsGait, ContainsGaitResponse, PossibleGaits, PossibleGaitsResponse,
                                         SetGaitVersion)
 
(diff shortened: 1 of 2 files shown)


march_gait_selection/src/march_gait_selection/gait_selection_node.py
```diff
@@ -65,6 +75,8 @@ def main():
 
     rospy.core.add_preshutdown_hook(lambda reason: gait_state_machine.request_shutdown())
 
+    rospy.Subscriber('/march/error', Error, lambda msg: error_cb(gait_state_machine, msg))
+
```
Roelemans, line 79: Is it logical to do this here instead of directly in the state machine?

Olavhaasie: I find this design quite elegant, since the State machine does not depend on ROS topics and can, therefore, be easily tested without that extra dependency. Also right now all services and topics are initialized in the main node.

Roelemans: Okay :+1: 
PS I don't really understand 
`Also right now all services and topics are initialized in the main node.`
There are also subscribers and publishers initialized in state_machine_input.py no?

Olavhaasie:
> There are also subscribers and publishers initialized in state_machine_input.py no?

Yes :upside_down_face: I guess I am not consistent with my own style, I'll see if I can fix this some other time. Perhaps when everything is done I have to make some small fixes to make it prettier.

PR #568, opened 2020-07-31
[gait-selection-develop] Feature: trajectory scheduling

Olavhaasie:
Closes PM-506

## Description
This PR adds new gait scheduling to the gait state machine. Actually, there is not much scheduling done, it is just immediately sent to the controller, which will abort previous goals. I figured it could be useful for error catching and keeping the controller loosely coupled from the state machine. For the scheduling to work, I had to add a goal time tolerance on the, so that it would accept goals that are a fraction too late. It still follows the trajectory nicely, it's just the timing that was a bit imprecise, between scheduling and the actual performing of the gait.

## Changes
* Add `TrajectoryScheduler` class
* Remove `march_gait_scheduler` package
* Fix `march.launch` to use the new gait selection only

## Testing
You have to check out several branches: 

* `fix/new-scheduler` on simulation
* `remove-gait-prefix` on monitor
* `feature/new-state-machine` on gait-files

And then simply launch the new simulation

    roslaunch march_launch march_simulation.launch


march_launch/launch/march.launch
```diff
@@ -47,12 +45,6 @@
     <node name="rviz" pkg="rviz" type="rviz" if="$(eval not headless and rviz)"/>
 
     <group ns="march">
-        <include file="$(find march_state_machine)/launch/state_machine.launch">
-            <arg name="state_machine_viewer" value="$(eval not headless and state_machine_viewer)"/>
-            <arg name="unpause" value="$(eval configuration == 'simulation' and unpause)"/>
```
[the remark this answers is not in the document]
Olavhaasie: I don't know if you have launched the simulations, but I found that disabling the paused startup was fine🙂

march_gait_selection/src/march_gait_selection/state_machine/gait_state_machine.py
```diff
@@ -110,7 +120,8 @@ def _process_gait_state(self, elapsed_time):
         trajectory, should_stop = self._current_gait.update(elapsed_time)
         # schedule trajectory if any
         if trajectory is not None:
-            rospy.loginfo('Received new trajectory to schedule: ' + str(trajectory))
+            rospy.loginfo('Received new trajectory to schedule')
```
JorisWeeda, line 123: I don't think this info statement is very informative, it gets published every execution of a subgait. I would either turn it into a _debug_ logging or add the trajectory name (subgait name) to the statement. You could also consider doing this for line 99.

Olavhaasie: My next PR will include publishing the current gait and subgait. I could add the subgait name in that one. Thanks for the suggestion, hadn't even thought of that yet :+1:

Olavhaasie:
> I've also noticed that you are not a fan of adding docstrings to your functions. I think you would save yourself a lot of question from the new march team if some documentation is added in the code.

I only add function docstrings to functions I think are not completely self explanatory or have to be implemented a lot (`GaitInterface` for example).

> When directly stopping the walk state after activating it the left and right swing are still executed. I don't think it is that big of a problem but are you aware of this behavior?

Yes I am aware of this. This has to do with how the walk gait graph is implemented. When you open with right_open and immediately press stop, the gait will remember that you pressed stop and will stop as soon as possible. This will happen in the next left_swing, which will transition to right_close. That is why there are 3 steps before stop, this has always been this way.

march_gait_selection/src/march_gait_selection/state_machine/gait_state_machine.py
JorisWeeda: > > I've also noticed that you are not a fan of adding docstrings to your functions. I think you would save yourself a lot of question from the new march team if some documentation is added in the code.
> 
> I only add function docstrings to functions I think are not completely self explanatory or have to be implemented a lot (`GaitInterface` for example).
> 
Hmm could be just me.
> > When directly stopping the walk state after activating it the left and right swing are still executed. I don't think it is that big of a problem but are you aware of this behavior?
> 
> Yes I am aware of this. This has to do with how the walk gait graph is implemented. When you open with right_open and immediately press stop, the gait will remember that you pressed stop and will stop as soon as possible. This will happen in the next left_swing, which will transition to right_close. That is why there are 3 steps before stop, this has always been this way.

Allright, nice

JorisWeeda:
> > I've also noticed that you are not a fan of adding docstrings to your functions. I think you would save yourself a lot of question from the new march team if some documentation is added in the code.
> 
> I only add function docstrings to functions I think are not completely self explanatory or have to be implemented a lot (`GaitInterface` for example).
> 
Hmm could be just me.
> > When directly stopping the walk state after activating it the left and right swing are still executed. I don't think it is that big of a problem but are you aware of this behavior?
> 
> Yes I am aware of this. This has to do with how the walk gait graph is implemented. When you open with right_open and immediately press stop, the gait will remember that you pressed stop and will stop as soon as possible. This will happen in the next left_swing, which will transition to right_close. That is why there are 3 steps before stop, this has always been this way.

Allright, nice

PR #571, opened 2020-08-04
[gait-selection-develop] Feature: current gait

Olavhaasie:
Closes PM-541

## Description
This PR adds the ability to publish current gaits and states. The two topics I have added are

1. `/march/gait_selection/current_state` publishes state name and type on every transition
2. `/march/gait_selection/current_gait` publishes info about the current trajectory being executed

These events are triggered by callbacks in the state machine. I have made it so that is it easy to add more such callbacks in the future, for example, playing sounds, which will be my next issue.

## Changes
* Add new message descriptions
* Add callbacks to state machines that get called on specific events.
* Changed the topics and removed all references to the old schedule topic

## Testing
Checkout `gait-selection-develop` on project-march/monitor and project-march/simulation and run the simulation.


march_gait_selection/src/march_gait_selection/state_machine/setpoints_gait.py
```diff
@@ -14,6 +14,34 @@ def __init__(self, gait_name, subgaits, graph):
     def name(self):
         return self.gait_name
 
+    @property
+    def subgait_name(self):
+        if self._current_subgait in self.subgaits:
+            return self.subgaits[self._current_subgait].subgait_name
+        else:
+            return None
+
```
RutgerVanBeek, line 23: The return None is not necessary or did you want to make it explicit?
(for all the properties of course)

Olavhaasie: Ah yes you're right these are not necessary. But I think it's good that they are explicit.

march_gait_selection/src/march_gait_selection/state_machine/gait_state_machine.py
```diff
@@ -184,3 +220,15 @@ def _generate_home_gaits(self, idle_positions):
...
+    def _add_callback(callbacks, cb):
+        """Adds a method to a list if it is callable."""
+        if callable(cb):
+            callbacks.append(cb)
+
+    @staticmethod
+    def _call_callbacks(callbacks, *args):
+        """Calls multiple methods with same set of arguments."""
+        for cb in callbacks:
+            cb(*args)
```
Roelemans, line 234: Why make this a separate method instead of just directly calling the callbacks in _call_transition_callbacks and _call_gait_callbacks?

march_gait_selection/src/march_gait_selection/state_machine/home_gait.py
```diff
@@ -21,6 +21,14 @@ def __init__(self, name, position, duration=3.0):
...
         return self._name
 
+    @property
+    def subgait_name(self):
+        return self._name
+
+    @property
+    def duration(self):
+        return self._duration
+
```
Roelemans, line 31: Hmm home gaits often have a gait type too right? Especially since gains are zero at startup, and only get changed when the first homing gait is performed

march_gait_selection/src/march_gait_selection/state_machine/gait_state_machine.py
Olavhaasie: Because then I would have to do this multiple times and I am planning on adding more callbacks, e.g. for sounds.

march_gait_selection/src/march_gait_selection/state_machine/home_gait.py
Olavhaasie: What do you suggest? Using the gait type that is used the most by gaits that come after it? Standing for example has a lot of gaits that are walk like, but some that are stairs like, so that should result in walk like, shouldn't it? I find that the default gait type is good enough for most home gaits and the gain scheduling will be replaced by the new controller anyway so..

Roelemans: I was thinking about the gait that comes before it, but this has the same problem that stairs up and walk both end in standing but have different gait types. It would however work for sitting, since that only has sit_down. This is also the main reason I suggest this - sitting should use the sit-like tuning. This is one of the mean reasons gain scheduling exists in the first place. So arguably it would be useful to use the most common gait type for a home gait?

march_gait_selection/src/march_gait_selection/state_machine/gait_state_machine.py
Roelemans: Hmm that would only be a single extra line (for the loop) in each of these, but I guess this is fine too.

march_gait_selection/src/march_gait_selection/state_machine/home_gait.py
Olavhaasie: I have made an even better fix. You can add the gait type to the position manually, so that you have always have the correct and same one.

PR #573, opened 2020-08-06
Increase the p and d value of the knees sjaan gait yaml

Roelemans:
## Description

In https://github.com/project-march/march/pull/570 new pid tuning for the FE's was introduced and tested for groundgaiting. The knee tuning is adopted in sjaan_gait.yaml in this PR. The HFE tuning was still weaker in groundgait yaml, hence they have not been changed. This should be tested in a training before merging.


RutgerVanBeek:
We tried this tuning at training, but it had the opposite effect. It didn't work, we should look into why, but at the short term for the reveal we should not change the tuning (this heavily).

PR #575, opened 2020-08-10
[gait-selection-develop] Feature: playing sounds

Olavhaasie:
Closes PM-508

## Description
Adds sound effects on the same actions as the old state machine did.

## Changes
* Add sound argument for gait selection
* Add sound client for playing sound and hooking into the state machine
* Remove the old state machine

## Testing
Checkout gait-selection-develop on project-march/monitor and project-march/simulation and run the simulation with `sounds:=true`


march_gait_selection/src/march_gait_selection/gait_selection_node.py
```diff
@@ -115,4 +116,20 @@ def current_gait_cb(gait_name, subgait_name, version, duration, gait_type):
     gait_state_machine.add_transition_callback(current_state_cb)
     gait_state_machine.add_gait_callback(current_gait_cb)
 
+    if rospy.get_param('~sounds', False):
+        sounds = Sounds(['start', 'gait_start', 'gait_end', 'gait_stop'])
+
+        def play_gait_sound(_state, is_idle):
+            if is_idle:
```
Roelemans, line 123: I find the is_idle name not very intuitive: does that mean it is idle before the gait executes or after? Maybe change the name to next_is_idle?

march_gait_selection/src/march_gait_selection/gait_selection_node.py
```diff
@@ -115,4 +116,20 @@ def current_gait_cb(gait_name, subgait_name, version, duration, gait_type):
...
+                sounds.play('gait_start')
+
+        gait_state_machine.add_transition_callback(play_gait_sound)
+        gait_state_machine.add_stop_accepted_callback(lambda: sounds.play('gait_stop'))
+
+        # Short sleep is necessary to wait for the sound topic to initialize
+        rospy.sleep(0.5)
+        sounds.play('start')
+
     gait_state_machine.run()
```
Roelemans, line 135: I usually have the feeling that node files are generally short and basically only launch the (several pieces) of code that do the actual work. This file is getting rather long with all these callback definitions. Do you think this would be better moved to a separate file? If you think it is fine like this I will adjust my worldview and not complain about it anymore :innocent:

Olavhaasie: I could split the `main` function up in multiple functions? To be honest I wouldn't know how to split it into multiple files, the sound library is already another file.

Olavhaasie: I have split up the `main` method into multiple methods, so now you can read it without having to scroll :+1:

PR #578, opened 2020-08-11
Feature: add controller launch arguments

Deakensbubbles:
…controllers i.e. /controller_type/march4.yaml

Closes PM-None existent but is still pretty nice for changing the controllers

## Description

adds a launch argument that allows the user to change the directory file in the simulation, and later also in the hardware interface, which contain the .yaml files for the controllers. This way we can easily run the simulation with effort control as well as with Gainscheduling 2.0

## Changes

Added launch argument: Controller which defaults to effort_control for the march.launch file as well as for the march_simulation.launch file.

### DIFF
--- a/march_launch/launch/march_simulation.launch
+++ b/march_launch/launch/march_simulation.launch
@@ -21,9 +21,12 @@
 
     <arg name="rosbag" default="false" doc="Enables rosbag record node."/>
 
+    <arg name="controller" default="effort_control" doc="Changes the controller used by simulation."/>
+
     <param name="balance" value="$(arg balance)"/>
 
     <include file="$(dirname)/march.launch" pass_all_args="true">
         <arg name="configuration" value="simulation"/>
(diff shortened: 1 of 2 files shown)


march_launch/launch/march_simulation.launch
```diff
@@ -21,9 +21,12 @@
 
     <arg name="rosbag" default="false" doc="Enables rosbag record node."/>
 
+    <arg name="controller" default="effort_control" doc="Changes the controller used by simulation."/>
+
     <param name="balance" value="$(arg balance)"/>
 
     <include file="$(dirname)/march.launch" pass_all_args="true">
         <arg name="configuration" value="simulation"/>
+        <arg name="controller" value="$(arg controller)"/>
```
Olavhaasie, line 30: This line is not necessary since `pass_all_args` is set to true, which forwards all args defined in the launch file.

PR #579, opened 2020-08-18
Use defined setters in constructor

ThijsRay:
## Description

Let the `Setpoint` and `JointTrajectory` classes use their own setters in their constructor. This removes code duplication.

## Changes

Replace assignments of variables in constructor of `Setpoint` and `JointTrajactory` with implicit calls to the setters of the variables.

### DIFF
--- a/march_shared_classes/src/march_shared_classes/gait/joint_trajectory.py
+++ b/march_shared_classes/src/march_shared_classes/gait/joint_trajectory.py
@@ -14,11 +14,10 @@ class JointTrajectory(object):
     def __init__(self, name, limits, setpoints, duration, *args):
         self.name = name
         self.limits = limits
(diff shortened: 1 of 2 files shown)


ThijsRay:
I agree that this behaviour might be overlooked, because it is less explicit. It should be clear for the reader what the code is doing. On the other hand, these setters are used outside of this class as well. Why should the constructor be treated differently?

Olavhaasie:
I agree with Thijs that this reduces code duplication and looks nicer. Although it might be harder to comprehend, I don't think that is an issue, since, as Thijs already said, it also happens outside the class.

Olavhaasie (approved):

Roelemans (approved):

PR #582, opened 2020-08-25
Feature/blc 48 capture point service call

RutgerVanBeek:
Closes BLC-48

## Description

I created a new PR, because a lot was done since the start of that PR and all comments and the description were outdated. Also credits to @JorisWeeda and @Roelemans who also did a lot of the work. 

In this PR the new inverted Pendulum based capture point calculation is used by the balance gait. There is now just one capture point calculating object (not relating to one of the feet anymore). A service call is implemented to request the capture point from the capture point calculator. This is then used to set a pose goal for the swing leg in the balance gait. Goals are now set via `JointStates` as the API seems to handle them better. 
## Changes

* Remove the dependency on a foot and tf for the capture point calculatore
* Implement a service call for requesting a capture point calculation
* Update the `BalanceGait` object the handle the new flow. 
* Check for the balance parameter in the gait selection, so no object is created when the parameter is false
## Testing
Checkout this branch and set the `balance:=true` when launching. Checkout `feature/start-balance-open-with-ankle-forward-movement` and `feature/turn-off-com-plugin` for the full falling experience.


march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -1,70 +1,110 @@
...
+        :param duration:
+            the amount of seconds away from the current time the capture point should be calculated
+        """
+        falling_time = InvertedPendulum.calculate_falling_time(
+            self._center_of_mass.x,
+            self._center_of_mass.y,
+            self._center_of_mass.z,
+            self.vx, self.vy)
+
+        capture_point_duration = min(duration, 0.5 * falling_time)
```
Joris-JW: why do we multiple the falling time with 0.5?

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -1,70 +1,110 @@
...
+
+        new_center_of_mass = InvertedPendulum.numeric_solve_to_t(
+            self._center_of_mass.x,
+            self._center_of_mass.y,
+            self._center_of_mass.z,
+            self.vx, self.vy, capture_point_duration)
+
+        if new_center_of_mass['z'] <= 0:
+            rospy.logdebug_throttle(1, 'Cannot calculate capture point; z of new center of mass <= 0')
```
Joris-JW: We state that the capture point cannot be calculated but we do move on (also in the `get_capture_point` we do not verify if this condition is true) maybe add a return statement? or create a boolean which states that it is verified?

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
```diff
@@ -69,41 +67,25 @@ def default_walk(self, new_default_walk):
...
 
-        if self._old_capture_point_msg_time[leg_name]:
-            if self._old_capture_point_msg_time[leg_name] == self._latest_capture_point_msg_time[leg_name]:
-                rospy.logwarn('No new capture point messages received from cp topic of {lg}'.format(lg=leg_name))
-                return None
-
-        self._old_capture_point_msg_time[leg_name] = self._latest_capture_point_msg_time[leg_name]
+        if not return_msg.success:
+            rospy.logwarn('No messages received from the capture point service.')
+            return None
```
Joris-JW: shouldn't we return the default subgait instead of none? Not sure what the behavior will be if it reaches this statement, will the movegroup do anything?

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
```diff
@@ -192,9 +179,11 @@ def construct_subgait(self, capture_point_leg_name, subgait_name):
             return self.default_walk[subgait_name]
 
         balance_trajectory_subgait = self.create_subgait_of_trajectory(balance_trajectory, subgait_name)
-        balance_trajectory_max_joint_duration = max(joint.duration for joint in balance_trajectory_subgait.joints)
+        balance_trajectory_duration = min(sg_duration, cp_duration)
```
Joris-JW: Im not so sure about this `min` anymore. Moveit! states that it would plan the trajectory as fast as the limits allow it to be . If this duration is larger compared to the subgait duration it should not be scaled otherwise the limits would be surpassed. The only thing is in practice we experience different behavior. Maybe add an extra comment, what do u guys think?

Joris-JW (commented):
I like what I see! nice cleaning and I think the method is much better implemented. Some general question, furthermore looks good. Approve is close :fire:

march_data_collector/src/march_data_collector/cp_calculator.py
RutgerVanBeek: I thought this was more of a heuristic, with the falling time you are too late. May be @Roelemans knows more about this.

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
RutgerVanBeek: I don't know exactly what was the idea here, but if the falling time is much larger then the gait time, I wouldn't go as fast as possible.

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
RutgerVanBeek: It would return a duration if successful, so I think I will return -1 (invalid duration)

march_data_collector/src/march_data_collector/cp_calculator.py
Roelemans: Indeed; The falling time is the time when the CoM hits the ground, so obviously you have to be faster than that. However, it might be neat to make it a constant; something like `FRACTION_OF_FALLING` or something

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
Roelemans: The sg_duration is the duration of a normal swing gait, not the duration of what MoveIt produces. I think this min is fine, at most you will want to check whether this result is smaller than the duration of the MoveIt gait, and throw an error if it is (because MoveIt should already be max speed). However, since MoveIt doesn't work yet, we can at most add a comment about this as you mentioned.

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -1,70 +1,113 @@
...
+
+        x_cp = new_center_of_mass['x'] + new_center_of_mass['vx'] * capture_point_multiplier
+        y_cp = new_center_of_mass['y'] + new_center_of_mass['vy'] * capture_point_multiplier
+
+        self._capture_point_marker.header.stamp = rospy.get_rostime()
+        self._capture_point_marker.pose.position.x = x_cp
+        self._capture_point_marker.pose.position.y = y_cp
+        self._capture_point_marker.pose.position.z = 0
+
+        self.cp_publisher.publish(self._capture_point_marker)
```
Roelemans: Why return the capture point duration while the capture point marker is written to a class variable? I think you should either return both or write both to a class variable

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -1,70 +1,113 @@
...
+            self._center_of_mass.z,
+            self.vx, self.vy)
+
+        capture_point_duration = min(duration, 0.5 * falling_time)
+
+        new_center_of_mass = InvertedPendulum.numeric_solve_to_t(
+            self._center_of_mass.x,
+            self._center_of_mass.y,
+            self._center_of_mass.z,
+            self.vx, self.vy, capture_point_duration)
```
Roelemans: You need the location of the CoM here relative to the stable foot. (i.e. the xyz coordinates in an axis system where the stable foot is at (0,0,0))

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
```diff
@@ -1,45 +1,40 @@
...
+        self._capture_point_service = rospy.ServiceProxy('/march/capture_point', CapturePointPose)
 
-        rospy.Subscriber('/march/cp_marker_foot_left', Marker, queue_size=1, callback=self.capture_point_cb,
-                         callback_args='left_leg')
-        rospy.Subscriber('/march/cp_marker_foot_right', Marker, queue_size=1, callback=self.capture_point_cb,
-                         callback_args='right_leg')
+        self._joint_state_target = JointState()
+        self._joint_state_target.header = Header()
+        self._joint_state_target.name = self.move_group['left_leg'].get_active_joints() + self.move_group['right_leg'].\
+            get_active_joints()
```
Roelemans:
```suggestion
        self._joint_state_target.name = self.move_group['left_leg'].get_active_joints() + 
                                                          self.move_group['right_leg'].get_active_joints()
```
better to split after/before a `+`

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
```diff
@@ -69,41 +67,25 @@ def default_walk(self, new_default_walk):
...
 
-        self.move_group[leg_name].set_joint_value_target(pose, end_effector, True)
-
-    def calculate_normal_trajectory(self, leg_name, subgait_name):
-        """Calculate the pose of the non-capture-point leg and set this as the pose for this leg.
+    def set_stance_leg_target(self, leg_name, subgait_name):
+        """Set the target of the stance leg to the end of the gait file.
 
         :param leg_name: The name of the move group which does not use capture point
         :param subgait_name: the normal subgait name
```
Roelemans, line 99: Add a a dostring about the return value here :)

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
```diff
@@ -171,14 +155,17 @@ def construct_subgait(self, capture_point_leg_name, subgait_name):
         """
         non_capture_point_move_group = 'right_leg' if capture_point_leg_name == 'left_leg' else 'left_leg'
 
-        self.calculate_capture_point_trajectory(capture_point_leg_name)
-        self.calculate_normal_trajectory(non_capture_point_move_group, subgait_name)
+        cp_duration = self.set_swing_leg_target(non_capture_point_move_group, subgait_name)
+        sg_duration = self.set_stance_leg_target(capture_point_leg_name, subgait_name)
```
Roelemans: Maybe a bit more descriptive name then sg? just write out stance_gait_duration?

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
Roelemans: I think it will be good to verify one last time whether the joint orders are still consistent. (once cp calculation is fixed ;))

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
```diff
@@ -192,9 +179,11 @@ def construct_subgait(self, capture_point_leg_name, subgait_name):
             return self.default_walk[subgait_name]
 
         balance_trajectory_subgait = self.create_subgait_of_trajectory(balance_trajectory, subgait_name)
-        balance_trajectory_max_joint_duration = max(joint.duration for joint in balance_trajectory_subgait.joints)
+        balance_trajectory_duration = min(sg_duration, cp_duration)
+
+        rospy.loginfo('Balance subgait with duration {dr}'.format(dr=str(balance_trajectory_duration)))
```
Roelemans:
```suggestion
        # Write produced subgait to file for easy evaluating
        rospy.loginfo('Balance subgait with duration {dr}'.format(dr=str(balance_trajectory_duration)))
```

march_shared_resources/srv/CapturePointPose.srv
```diff
@@ -0,0 +1,7 @@
+# Service for getting the position of the center of mass after falling for time duration
```
Roelemans:
```suggestion
# Service for getting the position of the capture point, after letting the center of mass falling for time duration (this time can be used to make a step)
```

march_shared_resources/srv/CapturePointPose.srv
```diff
@@ -0,0 +1,7 @@
...
+
+float64 duration
+---
+bool success
```
Roelemans, line 5: What is succes? Maybe make the name more descriptive? (e.g. "capture_point_exists")

march_gait_selection/src/march_gait_selection/gait_selection.py
```diff
@@ -24,11 +24,15 @@ def __init__(self, package, directory):
...
         self._robot = urdf.Robot.from_parameter_server('/robot_description')
-        self._balance_gait = BalanceGait.create_balance_subgait()
 
         self._gait_version_map = self._load_version_map()
         self._loaded_gaits = self._load_gaits()
 
+        is_balance_used = rospy.get_param('/balance', False)
+
+        if is_balance_used:
+            self._load_balance_gait()
```
Roelemans:
```suggestion
        is_balance_used = rospy.get_param('/balance', False)
        if is_balance_used:
            self._load_balance_gait()
```

Roelemans (changes requested):

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
RutgerVanBeek: I am pretty sure the order is consistent, because it is the order of the chain.

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
RutgerVanBeek: I will add a docstring to the method, I think that makes more sense.

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -72,35 +75,41 @@ def _calculate_capture_point(self, duration):
...
             the amount of seconds away from the current time the capture point should be calculated
         """
-        falling_time = InvertedPendulum.calculate_falling_time(
-            self._center_of_mass.x,
-            self._center_of_mass.y,
-            self._center_of_mass.z,
-            self.vx, self.vy)
+        try:
+            world_transform = self._tf_buffer.lookup_transform('world', self._static_foot_link, rospy.Time())
+            wt_translation = world_transform.transform.translation
```
Roelemans:
```suggestion
            static_foot_transform = self._tf_buffer.lookup_transform('world', self._static_foot_link, rospy.Time())
            static_foot_position = world_transform.transform.translation
```
I think something like this is more intuitive

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -72,35 +75,41 @@ def _calculate_capture_point(self, duration):
...
-        self._capture_point_marker.pose.position.y = y_cp
-        self._capture_point_marker.pose.position.z = 0
+            self._capture_point_marker.header.stamp = rospy.get_rostime()
+            self._capture_point_marker.pose.position.x = x_cp + wt_translation.x
+            self._capture_point_marker.pose.position.y = y_cp + wt_translation.y
+            self._capture_point_marker.pose.position.z = 0
+
+            self.cp_publisher.publish(self._capture_point_marker)
+        except tf2_ros.TransformException as e:
+            rospy.logdebug('Error in trying to lookup transform for capture point: {error}'.format(error=e))
```
Roelemans:
```suggestion
            rospy.loginfo(Error in trying to lookup transform for capture point: {error}'.format(error=e))
```

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -1,70 +1,126 @@
...
+
+            new_center_of_mass = InvertedPendulum.numeric_solve_to_t(
+                self._center_of_mass.x - wt_translation.x,
+                self._center_of_mass.y - wt_translation.y,
+                self._center_of_mass.z - wt_translation.z,
+                self.vx, self.vy, capture_point_duration)
+
+            if new_center_of_mass['z'] <= 0:
+                rospy.loginfo_throttle(1, 'Cannot calculate capture point; z of new center of mass <= 0')
+                return -1
```
Roelemans: Hmm you retrun -1 here, won't that error since you expect two outputs below?

march_shared_resources/srv/CapturePointPose.srv
Roelemans: :eyes:

Roelemans:
There is something wrong with the capture point calculation; the capture point of right_open_1 is placed to the left of the left leg, though it obviously shouldn't be there. We should have a look at this together I think

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
Roelemans: This is right_open_2
image
and this is the output of moveit
image
pretty sure that something is going wrong :s

march_shared_resources/srv/CapturePointPose.srv
RutgerVanBeek: Oopsie

march_data_collector/src/march_data_collector/cp_calculator.py
RutgerVanBeek: Ah, the two outputs is new, you are right.

march_shared_resources/srv/CapturePointPose.srv
RutgerVanBeek: After giving it some thought, I now think that succes describes what we mean. Did  the calculation succeed or not. `capture_point_exists` does not really cover it in my opinion as there are multiple reasons why the calculation may not have been successful, some unrelated to capture point. Let me know what you think

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
RutgerVanBeek: Found the bug and fixed it. Now the final positions do exactly correspond for the left leg. For the right leg it also looks quite similar. Nevertheless it is still falling of course :(

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -64,15 +65,11 @@ def center_of_mass_marker(self, updated_center_of_mass):
...
+        self.vx = (updated_center_of_mass.pose.position.x - self._center_of_mass.x) / self._delta_t
+        self.vy = (updated_center_of_mass.pose.position.y - self._center_of_mass.y) / self._delta_t
 
-        self.x = self._center_of_mass_marker.pose.position.x
-        self.y = self._center_of_mass_marker.pose.position.y
-        self.z = self._center_of_mass_marker.pose.position.z
-
-        self._center_of_mass_marker = copy.deepcopy(updated_center_of_mass)
-        self._prev_t = updated_center_of_mass.header.stamp
+        self._center_of_mass = updated_center_of_mass.pose.position
```
Roelemans:
```suggestion
        self._center_of_mass = copy.deepcopy(updated_center_of_mass.pose.position)
```
This deepcopy was here with a reason :innocent: (don't forget to add the import)

Roelemans (changes requested):

march_data_collector/src/march_data_collector/cp_calculator.py
```diff
@@ -1,70 +1,131 @@
...
+
+        self._gravity_constant = 9.81
+        self._prev_t = rospy.Time.now()
+        self._delta_t = 0
+
+        self.x = 0
+        self.y = 0
+        self.z = 0
+        self.vx = 0
+        self.vy = 0
```
Roelemans:
```suggestion
        self.com_vx = 0
        self.com_vy = 0
```
self.x,y,z are unused

Roelemans (changes requested):
Capture point is looking good now! :D I always get a `Error setting joint target. Is the target within bounds?` now though :cry:. We can just leave that as an issue for M6 though ;)

RutgerVanBeek:
> Capture point is looking good now! :D I always get a `Error setting joint target. Is the target within bounds?` now though . We can just leave that as an issue for M6 though ;)

Hmmm, may be if I have some time upcoming week I will look into it.

Roelemans (changes requested):
Good work :)

Roelemans (dismissed):
When you try so hard... :musical_note:

march_shared_resources/srv/CapturePointPose.srv
Joris-JW: I agree with Rutger.

Joris-JW (dismissed):
:fire: nice work boys

Roelemans (commented):
I always get a "Error setting joint target. Is the target within bounds?" :cry: Dunno what could be the problem :s

Roelemans (approved):
Whoopsie forgot to pull. Seems to work properly! Good job :)

Joris-JW (approved):
56 commits = :fire:

PR #583, opened 2020-08-25
Gait selection develop

Olavhaasie:
Closes PM-474

## Description
This is the *very big* PR that was coming for a long time now. Every change in this PR have been approved as separate steps, but it's nice to test everything together.

The build is unfortunately failing, because it uses the develop version of the downstream packages: monitor and simulation, which do not have the necessary changes.

## Changes
Some of the most important changes
* Removed `march_state_machine` and `march_gait_scheduler` package in favor of integrating those functionalities into the gait selection
* Gait state machine is generated on startup
* Gaits use graph descriptions in their configuration file to describe behavior
* No more intermediate action servers, instead 2 new topics
  - `CurrentGait` publishes details about every performed gait/subgait
  - `CurrentState` publishes the current state the state machine is in
* Gaits can now implement a `GaitInterface` to perform gaits instead of only preprogrammed trajectories

## Related PRs
* https://github.com/project-march/monitor/pull/68
* https://github.com/project-march/simulation/pull/63
* https://github.com/project-march/gait-files/pull/212

## Testing
Checkout the other PRs and start experimenting


march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
```diff
@@ -191,22 +159,68 @@ def construct_subgait(self, capture_point_leg_name, subgait_name):
...
+    @property
+    def starting_position(self):
+        return self._default_walk.starting_position
+
+    @property
+    def final_position(self):
+        return self._default_walk.final_position
+
+    def start(self):
+        self._current_subgait = self._default_walk.graph.start_subgaits()[0]
```
RutgerVanBeek, line 204: Shouldn't you set the duration here as well.

march_gait_selection/src/march_gait_selection/state_machine/gait_interface.py
```diff
@@ -0,0 +1,68 @@
+class GaitInterface(object):
+    @property
+    def name(self):
+        """Returns the name of the gait."""
+        return None
+
+    @property
+    def subgait_name(self):
+        """Returns the name of the currently executing trajectory."""
+        return None
```
RutgerVanBeek, line 10:
```suggestion
        """Returns the name of the currently executed trajectory."""
        return None
```

march_gait_selection/src/march_gait_selection/state_machine/gait_state_machine.py
```diff
@@ -0,0 +1,260 @@
...
+            self._is_idle = True
+            self._current_gait.end()
+            self._input.gait_finished()
+            self._call_transition_callbacks()
+            rospy.loginfo('Finished gait `{0}`'.format(self._current_gait.name))
+            self._current_gait = None
+
+    def _handle_input(self):
+        if self._input.stop_requested() or self._should_stop:
+            self._should_stop = False
```
RutgerVanBeek, line 168: Shouldn't you keep this True if stop is rejected?

march_gait_selection/src/march_gait_selection/state_machine/state_machine_input.py
```diff
@@ -0,0 +1,113 @@
...
+
+        :returns TransitionRequest
+        """
+        if self._transition_index == GaitInstruction.INCREMENT_STEP_SIZE:
+            return TransitionRequest.INCREASE_SIZE
+
+        if self._transition_index == GaitInstruction.DECREMENT_STEP_SIZE:
+            return TransitionRequest.DECREASE_SIZE
+
+        return TransitionRequest.NONE
```
RutgerVanBeek, line 41: Aren't you always returning the same integer here? Or is this for modularity?

march_gait_selection/src/march_gait_selection/state_machine/trajectory_scheduler.py
```diff
@@ -0,0 +1,30 @@
...
+        self._failed = False
+        self._trajectory_client = actionlib.SimpleActionClient(topic, FollowJointTrajectoryAction)
+
+    def schedule(self, trajectory):
+        """Schedules a new trajectory.
+
+        :param JointTrajectory trajectory: a trajectory for all joints to follow
+        """
+        self._failed = False
+        goal = FollowJointTrajectoryGoal()
```
RutgerVanBeek, line 17: It is nice to also set the stamp

RutgerVanBeek (changes requested):
Mainly just questions

RutgerVanBeek:
Screenshot from 2020-08-25 20-05-12
This happend while doing tilted path left start. However I could not reproduce this. 

Furthermore there is a lot more info in the terminal, what do we think about that, is it too spammy? @Roelemans @JorisWeeda

march_gait_selection/src/march_gait_selection/state_machine/state_machine_input.py
Olavhaasie: I don't really understand your question. It returns an enum, which depends on the transition index, when there is no transition it always returns `NONE`

march_gait_selection/src/march_gait_selection/state_machine/gait_state_machine.py
Olavhaasie: No otherwise it will still be True when the gait ends and you start a new gait. Furthermore, a gait rejects stop when it is not possible to stop the gait at all.

march_gait_selection/src/march_gait_selection/dynamic_gaits/balance_gait.py
Olavhaasie: Oopsie this quite a big mistake and I have no idea how this actually works

march_gait_selection/src/march_gait_selection/state_machine/trajectory_scheduler.py
Olavhaasie: The stamp in this message is used for planning the trajectory, when I add the current time here it will complain that the time of the first point is in the past. So it should be set to 0.

JorisWeeda:
> Furthermore there is a lot more info in the terminal, what do we think about that, is it too spammy? @Roelemans @JorisWeeda

I don't really mind the extra information. Compared to the old state machine it only has some additional information at the start and end of the gait (requested, accepted, executed).

RutgerVanBeek (dismissed):

Roelemans (changes requested):
I have found two flaws:
1) You are able to change versions of a gait while that gait is being executed. This results in a midgait transition from, for example, "MV_walk_leftswing_v2" to "MIV_final" version of right swing, even though these subgait versions do not necessarily connect.
2) You are able to change versions such that gait transitions are no longer correct. This is possbile with both the bend_knee_left_tilted_path gait and the walk gait (when I created a new version of right open where the start did not match with standing, it allowed me to change it and execute it without problems)  

Ideally, to solve 1), you would allow changing only subgait versions of the gaits that are not currently being executed. However, you can also just disallow all version changes while not idle.

Olavhaasie:
> You are able to change versions of a gait while that gait is being executed. This results in a midgait transition from, for example, "MV_walk_leftswing_v2" to "MIV_final" version of right swing, even though these subgait versions do not necessarily connect.

Fixed this as you suggested by disallowing changing subgait versions of active gaits.

> You are able to change versions such that gait transitions are no longer correct. This is possbile with both the bend_knee_left_tilted_path gait and the walk gait (when I created a new version of right open where the start did not match with standing, it allowed me to change it and execute it without problems)

This is very good that you noticed this, since I forgot to implement checking changing subgait versions that are connected to `start` or `end`. I have now added two extra checks that make sure the new subgait version has the same positions, so that the state machine does not have to be re-generated. In the future you could of course build a way of notifying the state machine of such a change and re-generate the state machine.

march_shared_classes/src/march_shared_classes/gait/gait.py
```diff
@@ -115,14 +119,30 @@ def set_subgait_versions(self, robot, gait_directory, version_map):
...
+                        raise NonValidGaitContent(
+                            'The starting position of new version {gait} {subgait} does not match'.format(
+                                gait=self.gait_name, subgait=to_subgait_name))
+                elif to_subgait_name == self.graph.END:
+                    old_subgait = self.subgaits[from_subgait_name]
+                    new_subgait = new_subgaits[from_subgait_name]
+
+                    if old_subgait.final_position != new_subgait.final_position:
+                        raise NonValidGaitContent(
+                            'The final position of new version {gait} {subgait} does not match'.format(
```
Roelemans:
```suggestion
                            msg='The final position of new version {gait} {subgait} does not match'.format(
```

march_shared_classes/src/march_shared_classes/gait/gait.py
```diff
@@ -115,14 +119,30 @@ def set_subgait_versions(self, robot, gait_directory, version_map):
...
-                                                  'not match'.format(gait=self.gait_name,
-                                                                     sn=from_subgait.subgait_name,
-                                                                     ns=to_subgait.subgait_name))
+                if from_subgait_name == self.graph.START:
+                    old_subgait = self.subgaits[to_subgait_name]
+                    new_subgait = new_subgaits[to_subgait_name]
+
+                    if old_subgait.starting_position != new_subgait.starting_position:
+                        raise NonValidGaitContent(
+                            'The starting position of new version {gait} {subgait} does not match'.format(
```
Roelemans:
```suggestion
                            msg='The starting position of new version {gait} {subgait} does not match'.format(
```

Roelemans (changes requested):
Nice job! One last thing :innocent: 
image

march_shared_classes/src/march_shared_classes/gait/subgait.py
```diff
@@ -298,6 +298,16 @@ def get_joint_names(self):
         """Get the names of all the joints existing in the joint list."""
         return [joint.name for joint in self.joints]
 
+    @property
+    def starting_position(self):
+        """Returns a dictionary of joint positions at the start of this subgait."""
+        return dict([(joint.name, joint.setpoints[0].position) for joint in self.joints])
```
RutgerVanBeek:
```suggestion
        return {joint.name: joint.setpoints[0].position for joint in self.joints}
```
It think this is the prettier way to do it.

march_shared_classes/src/march_shared_classes/gait/subgait.py
```diff
@@ -298,6 +298,16 @@ def get_joint_names(self):
...
+
+    @property
+    def final_position(self):
+        """Returns a dictionary of joint positions at the end of this subgait."""
+        return dict([(joint.name, joint.setpoints[-1].position) for joint in self.joints])
```
RutgerVanBeek:
```suggestion
        return {joint.name: joint.setpoints[-1].position for joint in self.joints}
```

RutgerVanBeek (changes requested):
Is it on purpose that the time is not automatically running, when starting? I would prefer it do so.

Olavhaasie:
> Is it on purpose that the time is not automatically running, when starting? I would prefer it do so.

Do you mean that the simulation is paused? Because that should not be the default

Roelemans (approved):

RutgerVanBeek:
> > Is it on purpose that the time is not automatically running, when starting? I would prefer it do so.
> 
> Do you mean that the simulation is paused? Because that should not be the default

Last time I also wanted to test balance and checked out Roel's turn-off-com-branch, because of that didn't default to unpause. So it actually was allright when I check out the right branch in simulation now.

RutgerVanBeek (approved):

PR #585, opened 2020-08-27
Add simple check for existence of gait directory

KatjaSchmahl:
Closes PM-235

## Description

Check whether the gait directory exists and give an extra warning that is more specific than FileNotFound of default.yaml.
## Testing
A test for this fix was already present


Joris-JW (changes requested):
It would be nice to add a small test to verify the exception.

JorisWeeda:
> It would be nice to add a small test to verify the exception.

Joris-JW (approved):
> It would be nice to add a small test to verify the exception.

I've noticed that there are indeed some tests which already verify the FileNotFoundError & PackageNotFoundError exception. So the comment above can be ignored.

Roelemans (approved):

PR #586, opened 2020-09-13
Feature: Update pid gains when different

RutgerVanBeek:
No issue for this
## Description

Previously the interpolation was only started when a gait with a different gait type was selected. I have improved upon this by updating whenever the current and desired are not equal. This is beneficial when you update the parameters during runtime. When you select the next gait an interpolation towards the new values is done even when the gait type is equal. 

## Changes

* Create a function to test whether interpolation is necessary. 
* Check whether the needed gains are equal to current gains before interpolating
## Testing

One should launch the simulation with the parameter `gain_scheduling` set equal to true. One can update the parameter values with `rosparam load [path to valid yaml] /march/gain_scheduling_node`. It should start interpolating even when the gait types are equal.


march_gain_scheduling/src/march_gain_scheduling/dynamic_pid_reconfigurer.py
```diff
@@ -27,34 +27,33 @@ def gait_selection_callback(self, data):
...
-            self.interpolation_done = False
+        self._gait_type = new_gait_type
+        self.load_current_gains()
+        needed_gains = [self.look_up_table(i) for i in range(len(self._joint_list))]
+        rospy.logwarn(needed_gains)
+        self.done_interpolation_test(needed_gains)
 
-            needed_gains = [self.look_up_table(i) for i in range(len(self._joint_list))]
+        if not self.interpolation_done:
+            self.load_current_gains()
```
Roelemans:
```suggestion
```
This seems redundant

march_gain_scheduling/src/march_gain_scheduling/dynamic_pid_reconfigurer.py
```diff
@@ -27,34 +27,33 @@ def gait_selection_callback(self, data):
...
-            self.load_current_gains()
-            self.interpolation_done = False
+        self._gait_type = new_gait_type
+        self.load_current_gains()
+        needed_gains = [self.look_up_table(i) for i in range(len(self._joint_list))]
+        rospy.logwarn(needed_gains)
+        self.done_interpolation_test(needed_gains)
 
-            needed_gains = [self.look_up_table(i) for i in range(len(self._joint_list))]
+        if not self.interpolation_done:
```
Roelemans: self.interpolation_done no longer needs to be a class variable, as it is only used within this method.
```suggestion
        if not interpolation_done:
```

Roelemans (changes requested):

march_gain_scheduling/src/march_gain_scheduling/dynamic_pid_reconfigurer.py
```diff
@@ -10,7 +10,7 @@ class DynamicPIDReconfigurer:
     def __init__(self, joint_list):
         self._gait_type = None
         self._joint_list = joint_list
-        self.interpolation_done = True
+        # self.interpolation_done = True
```
Roelemans:
```suggestion
```
:eyes:

Roelemans (changes requested):

march_gain_scheduling/src/march_gain_scheduling/dynamic_pid_reconfigurer.py
RutgerVanBeek: Oh wow

march_gain_scheduling/src/march_gain_scheduling/dynamic_pid_reconfigurer.py
```diff
@@ -27,34 +26,30 @@ def gait_selection_callback(self, msg):
...
-            self._gait_type = new_gait_type
-            self.load_current_gains()
-            self.interpolation_done = False
+        self._gait_type = new_gait_type
+        self.load_current_gains()
+        needed_gains = [self.look_up_table(i) for i in range(len(self._joint_list))]
+        interpolation_done = self.done_interpolation_test(needed_gains)
 
-            needed_gains = [self.look_up_table(i) for i in range(len(self._joint_list))]
+        if not interpolation_done:
```
Roelemans:
```suggestion
        if not self.done_interpolation_test(needed_gains):
```

march_gain_scheduling/src/march_gain_scheduling/dynamic_pid_reconfigurer.py
```diff
@@ -27,34 +26,30 @@ def gait_selection_callback(self, msg):
...
             new_gait_type = 'walk_like'
 
-        if self._gait_type != new_gait_type or not self.interpolation_done:
-            self._gait_type = new_gait_type
-            self.load_current_gains()
-            self.interpolation_done = False
+        self._gait_type = new_gait_type
+        self.load_current_gains()
+        needed_gains = [self.look_up_table(i) for i in range(len(self._joint_list))]
+        interpolation_done = self.done_interpolation_test(needed_gains)
```
Roelemans:
```suggestion
```

Roelemans (changes requested):

Roelemans (approved):

Olavhaasie (approved):

Joris-JW (approved):
:fire:

PR #587, opened 2020-10-08
Fix double balance definition bug

Roelemans:
## Description

This removes a redundant definition of the balance gait in the gait_selection_node. This definition caused an error as it was also called when balance:=false. The proper definition is in GaitSelection._load_balance_gait() which is called from its __init__.


Olavhaasie (changes requested):
This was meant to be this way :see_no_evil:, you must have merged the incorrect change. I vote for removing the one in `GaitSelection`

Olavhaasie (approved):

RutgerVanBeek (approved):

PR #589, opened 2020-10-29
Feature/increase zero position ankles

Roelemans:
## Description

The new shoes have a much less flexible sole. Because of this, the ankles need to bend forward more to compensate. The easiest way to this is by changing  zero positions by modifying the values in the URDF. This PR increases the zero positions by 4 degrees.

This has the downside that some gaits may have target positions outside of the soft limits. This will be cut off by the safety controller, but it is not ideal. We tested last Tuesday if it was possible to push the ankle through its soft error limits, but Nadine could only give a third of the necessary force by putting her entire weight onto exo.


RutgerVanBeek:
Looks fine, although I think you want to merge into master.

